PR #1, opened 2022-06-25
Add Quilt support

leo60228:
This was based on the Fabric support, with the necessary changes made.

### DIFF
--- a/README.org
+++ b/README.org
@@ -38,10 +38,6 @@ From there, you can setup the service or use the packages, as described below.
 This README file is incomplete, and doesn't fully describe the ~services.minecraft-servers~ module.
 Additionally, documentation should be added for the maintenance of the ~vanillaServers.*~ and ~fabricServers.*~.
 
-** TODO: Package Quilt
-
-Should be similar to Fabric.
-
 * Packages
 
 ** ~vanillaServers.*~
@@ -72,13 +68,14 @@ fabricServers.fabric-22w16b
 fabricServers.fabric-1_18_2-0_13_3 # Specific fabric loader version
 #+end_src
 
-** TODO: ~quiltServers.*~
+** ~quiltServers.*~
+[[./pkgs/quilt-servers][Source]]
 
-Not yet present, but in the future the ~quiltServers~ will function the same as ~fabricServers~, but with the Quilt mod loader.
+~quiltServers~ functions the same as ~fabricServers~, but with the Quilt mod loader.
 
 ** ~minecraftServers.*~
 
-~vanillaServers // fabricServers~. Will be used most often as it contains all of the different server versions across each mod loader. When using the overlay, this will replace the Nixpkgs ~minecraftServers~.
+~vanillaServers // fabricServers // quiltServers~. Will be used most often as it contains all of the different server versions across each mod loader. When using the overlay, this will replace the Nixpkgs ~minecraftServers~.
 
 ** Others
 
--- a/pkgs/quilt-servers/default.nix
+++ b/pkgs/quilt-servers/default.nix
@@ -1,0 +1,40 @@
+{ callPackage
+, lib
+, vanillaServers
+}:
+
+let
+  versions = lib.importJSON ./locks.json;
+
+  inherit (lib.our) escapeVersion;
+  latestVersion = escapeVersion (lib.our.latestVersion versions);
+
+  packages =
+    builtins.foldl' (x: y: x // y) { }
+      (lib.mapAttrsToList
+        (lversion: gversions:
+          lib.mapAttrs'
+            (gversion: lock:
+              lib.nameValuePair
+                "quilt-${escapeVersion gversion}-${escapeVersion lversion}"
+                (callPackage ./server.nix { inherit lock; minecraft-server = vanillaServers."vanilla-${escapeVersion gversion}"; }))
+            gversions)
+        versions);
+in
+lib.recurseIntoAttrs (
+  packages
+  // (
+    let
+      removeVanilla = n: escapeVersion (lib.removePrefix "vanilla-" n);
+    in
+    (lib.mapAttrs'
+      (n: v: lib.nameValuePair "quilt-${lib.removePrefix "vanilla-" n}" v)
+      (lib.genAttrs
+        (builtins.filter
+          (n: (lib.hasPrefix "vanilla-" n) && (builtins.hasAttr "quilt-${removeVanilla n}-${latestVersion}" packages))
+          (builtins.attrNames vanillaServers))
+        (gversion: builtins.getAttr "quilt-${removeVanilla gversion}-${latestVersion}" packages)))
+  ) // {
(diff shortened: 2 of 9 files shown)


Infinidoge:
Thank you very much for the contribution!

If possible, please change the commit message to something along the lines of `feat(quilt-servers): init`, since that would better fit with the conventions I've used so far.

pkgs/quilt-servers/update.py
```diff
@@ -0,0 +1,150 @@
...
+
+import json
+import subprocess
+import requests
+import re
+from pathlib import Path
+
+
+def versiontuple(v):
+    return tuple(map(int, (v.partition("-")[0].split("."))))
```
Infinidoge, line 12: What particular cases in the API motivated changing this from `+` to `-`? I assume for parsing `1.18.2-pre1`, etc?

The `+` filtering in there was originally to remove the build number information from version numbers, so perhaps this should be updated to be consistent with the Fabric version.

(I might extract the common functions between the `update.py` files, but for now leaving them in their own files is fine; Makes it easy to be self-contained.)

pkgs/quilt-servers/server.nix
```diff
@@ -0,0 +1,13 @@
...
+, minecraft-server
+, jre_headless
+, lock
+}:
+let
+  loader = callPackage ./loader.nix { inherit lock; };
+in
+writeShellScriptBin "minecraft-server" ''
+  echo "serverJar=${minecraft-server}/lib/minecraft/server.jar" >> quilt-server-launcher.properties
+  exec ${jre_headless}/bin/java $@ -jar ${loader} nogui''
```
Infinidoge, line 13: I assume the `log4j` XML file not necessary for Quilt?

pkgs/quilt-servers/loader.nix
```diff
@@ -0,0 +1,39 @@
...
+      unzip -o $i
+    done
+
+    cat > META-INF/MANIFEST.MF << EOF
+    Manifest-Version: 1.0
+    Main-Class: org.quiltmc.loader.impl.launch.server.QuiltServerLauncher
+    Name: org/objectweb/asm/
+    Implementation-Version: 9.2
+    EOF
+  '';
```
Infinidoge, line 30: Roughly here is where in the Fabric file, there is
```nix
cat > fabric-server-launch.properties << EOF
launch.mainClass=${lock.mainClass}
EOF
```
I assume that this is unnecessary for Quilt?

pkgs/quilt-servers/default.nix
```diff
@@ -0,0 +1,42 @@
+{ callPackage
+, lib
+, vanillaServers
+}:
+
+let
+  versions = lib.importJSON ./locks.json;
+
+  merge = null;
```
Infinidoge, line 9: I honestly have no idea why I included this in the Fabric `default.nix`. Probably had a function there but forgot to remove it.

Feel free to remove it, and add an additional commit removing it from the Fabric side, as it (theoretically, hopefully...) is unimportant.

pkgs/quilt-servers/update.py
```diff
@@ -0,0 +1,150 @@
...
+)
+
+# Package all game versions supported by Quilt
+def GAME_VERSION_FILTER(version):
+    snapshotmatch = re.fullmatch('([0-9]{2})w([0-9]{1,2})[a-z]+', version["version"])
+
+    if snapshotmatch == None:
+        return versiontuple(version["version"]) >= (1, 18, 2)
+    else:
+        return tuple(map(int, snapshotmatch.groups())) >= (22, 11)
```
Infinidoge, line 30: Is there a particular reason why the game versions are constrained to `>= 1.18.2` and `>= 22w11x`?

I would assume that, though maybe not the best quality, all versions reported by the Quilt API would/should be supported, similar to what is done in the Fabric `update.py`

pkgs/quilt-servers/update.py
```diff
@@ -0,0 +1,150 @@
...
+# These filters specify which Quilt loader and Minecraft game versions to package.
+
+# Only package Quilt versions greater than 0.17.0 (using QuiltServerLauncher main class)
+LOADER_VERSION_FILTER = lambda version: (
+    version["separator"] == "." and versiontuple(version["version"]) >= (0, 17, 0)
+)
+
+# Package all game versions supported by Quilt
+def GAME_VERSION_FILTER(version):
+    snapshotmatch = re.fullmatch('([0-9]{2})w([0-9]{1,2})[a-z]+', version["version"])
```
Infinidoge, line 25: This regular expression should be declared up top as a constant with `re.compile`, as doing it like this will recompile the regular expression on each use, slowing things down slightly.

Infinidoge (changes requested):
Overall looks good! Please run `update.py` through `black` though, as it would help with keeping a consistent code style. (Note to self, make a `CONTRIBUTING.{md,org}`)

README.org
```diff
@@ -72,13 +68,13 @@ fabricServers.fabric-22w16b
 fabricServers.fabric-1_18_2-0_13_3 # Specific fabric loader version
 #+end_src
 
-** TODO: ~quiltServers.*~
+** ~quiltServers.*~
```
Infinidoge, line 71: Please also add a `Source` link here, like the Fabric and Vanilla server sections.

pkgs/quilt-servers/update.py
leo60228: Quilt uses version numbers like `0.16.0-beta.2`.

pkgs/quilt-servers/loader.nix
leo60228: Yep, the Quilt installer doesn't create it and it didn't seem to be necessary during testing.

pkgs/quilt-servers/update.py
leo60228: 1.18.2 was the current version as of Quilt's initial beta release, and I'm not sure how far back quilt-loader will still function. (Additionally, I didn't want to wait for every version to download)

pkgs/quilt-servers/server.nix
leo60228: 1.18.2+ already have secure log4j versions, don't they?

Infinidoge:
Another small thing I noticed, could you add to the README to change 
> along with versions for the Fabric loader

into
> along with versions for the Fabric and Quilt loaders

?

pkgs/quilt-servers/update.py
Infinidoge: Ah, I see, makes sense.

pkgs/quilt-servers/server.nix
Infinidoge: That is true. I don't know enough about log4j to know everything about what the file does, but I assume that the `Appenders` section is to recreate Vanilla behavior, while the `Loggers` section fixes the log4j exploit?

imsofi:
Friendly poke to see if this is still being worked on, i can help with adding some commits to this if help is needed.

Infinidoge:
Superseded by #7

PR #8, opened 2022-08-22
packwiz: remove packwiz from packages

imsofi:
Change reason: Packwiz is now properly packaged in nixpkgs, there is no
longer a reason to be packaging it inside of nix-minecraft.


imsofi:
Link to it in nixpkgs: https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/tools/games/minecraft/packwiz/default.nix

Infinidoge:
Packwiz is also itself now a flake, so people can just retarget to the flake.
(I packaged packwiz for nixpkgs, and helped in the discussions around making it a flake lol)

Should have probably removed it earlier.

Issue #6, opened 2022-08-02
Add examples


drestrepom:
Add examples of how to create a server and start it

drestrepom:
I am new to the world of nix, but I want to create a server based on nix to play with my friends

Titaniumtown:
I need examples too. I am a bit lost. I'll figure it out eventually though :p

sofiedotcafe:
I can probably look into this! :3

ahoneybun:
I'll second this as well, I'm trying to use the overlay but it still loads the version from nixpkgs instead.

PR #11, opened 2022-08-27
Full flakes coverage

soupglasses:
Currently our use of `outputs.packages` breaks nix due to design limitations to have everything on the top level be derivations and derivations only.

This is a draft PR as there are still work to be done:

- [x] Clean up how `missing commit hashes` was fixed. This one should be easy.
- [x] Formalize further which fabric versions we want to ship, and which we may skip due to lacking vanilla binaries.
- [x] Put test suites into fabric and quilt derivations so running `passthru.tests` will operate correctly.
- [ ] #6  Through `outputs.templates` and some guidance in the README.

I also recommend reading the commits themselves, as i have put time into writing comments for my reasoning of change there.


README.md
```diff
@@ -21,85 +19,72 @@ In your system configuration:
...
+There is also possibility of using this repo in a non-flake thanks to `flake-compat` using its outputs for `packages.<system>.PACKAGE`, `nixosModules.nix-minecraft`, and similar.
 
 ## Packages
 
-### `vanillaServers.*`
+### Vanilla servers
 [Source](./pkgs/minecraft-servers)
 
-An attrset of all of the vanilla server versions, in the form of `vanilla-version`, where `version` is the Minecraft version (`1.18`, `1.12.2`, `22w16b`, etc), with all periods and spaces replaced with underscores (`1_18`, `1_12_2`, etc).
+This repo contains all the vanilla server versions, in the form of `vanilla-version`, where `version` is the Minecraft version (`1.18`, `1.12.2`, `22w16b`, etc), with all periods and spaces replaced with underscores (`1_18`, `1_12_2`, etc).
```
Infinidoge, line 39: Nitpick: I think "all of the" sounds better here.
Looking it up I am apparently in the minority on that.
"All the" just sounds clunky to me, not sure why.

See also "All [versions] of the [fabric loader]" below.

README.md
```diff
@@ -21,85 +19,72 @@ In your system configuration:
...
 
-`quiltServers` functions the same as `fabricServers`, but with the Quilt mod loader.
+### Fabric servers
+[Source](./pkgs/fabric-servers)
 
-### `minecraftServers.*`
+This repo contains all versions of the fabric loader, in the form of `fabric-mcversion` and `fabric-mcversion-fabricversion`, following the same format as vanilla servers for its version numbers. If the `fabricversion` isn't specified, it will use the latest version.
 
-`vanillaServers // fabricServers // quiltServers`. Will be used most often as it contains all of the different server versions across each mod loader. When using the overlay, this will replace the Nixpkgs `minecraftServers`.
+The `mcversion` must be `>=1.14`, and if specified, the `fabricversion` must be `>=0.10.7`. The former is a limitation of Fabric, while the latter is the constraint put on the packaging lockfile to avoid exponential growth.
```
Infinidoge, line 60: (See main review comment)

This can probably be bumped up to at minimum Fabric version `0.13.X`.

README.md
```diff
@@ -21,85 +19,72 @@ In your system configuration:
...
 
-### Others
+For convenience, `fabric` is equivalent to the latest major Minecraft and Fabric loader versions.
```
Infinidoge, line 62:
```suggestion
For convenience, `fabric` is equivalent to the latest stable Minecraft and Fabric loader versions.
```

Correcting all of my mistakes in your PRs, apparently.

README.md
```diff
@@ -137,8 +122,6 @@ Would symlink a file containing `"Some text"` into the server's folder.
 
 This option is quite powerful, and can be used for a number of things, though most notably it can be used for declaratively setting up mods or plugins for the server.
 
-This example takes an attrset of the IDs and hashes for Modrinth mods, fetches each one, and makes a folder containing those mods. (`linkFarmFromDrvs` is quite useful because it can take a list of derivations and produce a folder suitable for this purpose.) The names in this attrset are meaningless, I only included them as convenient labels.
```
Infinidoge, line 140: Should probably keep something that explains `linkFarmFromDrvs` somewhere, since it's quite useful and in my mind, not inherently obvious what it does.

flake.nix
```diff
@@ -1,50 +1,35 @@
...
-          quilt-server = quiltServers.quilt;
-          minecraft-server = vanilla-server;
-        } // (
-          builtins.mapAttrs (n: v: callPackage v) (self.lib.rakeLeaves ./pkgs/helpers)
-        ) // (
-          builtins.mapAttrs (n: v: callPackage v { }) (self.lib.rakeLeaves ./pkgs/tools)
-        );
+        vanillaServers
+        // import ./pkgs/fabric-servers { our = self; inherit vanillaServers; inherit (pkgs) lib callPackage; }
+        // import ./pkgs/quilt-servers { our = self; inherit vanillaServers; inherit (pkgs) lib callPackage; }
```
Infinidoge, line 22: For clarity, I think it would be better to `s/our/flake/` everywhere.

flake.nix
```diff
@@ -1,50 +1,35 @@
...
-    , nixpkgs
-    , flake-utils
-    , ...
-    }@inputs:
+  outputs = { self, nixpkgs, ... }@inputs:
     let
-      packages = pkgs:
+      supportedSystems = [
+        "aarch64-linux"
+        "aarch64-darwin"
```
Infinidoge, line 10: Nitpick: 
```suggestion
        "aarch64-darwin"
        "aarch64-linux"
```
Sorted alphabetically.

pkgs/fabric-servers/default.nix
```diff
@@ -21,20 +22,17 @@ let
...
-          (builtins.attrNames vanillaServers))
-        (gversion: builtins.getAttr "fabric-${removeVanilla gversion}-${latestVersion}" packages)))
-  ) // {
-    fabric = builtins.getAttr "fabric-${escapeVersion vanillaServers.vanilla.version}-${latestVersion}" packages;
-  }
-)
+packages // (
+	let
+		removeVanilla = n: escapeVersion (lib.removePrefix "vanilla-" n);
+	in
```
Infinidoge, line 28: Should honestly just add `removeVanilla` to `./lib/default.nix`, since it gets used in both Fabric and Quilt, since then it can be brought in with `escapeVersion`.
Looks a little bit better, but in the end doesn't matter.

pkgs/fabric-servers/update.py
```diff
@@ -35,6 +37,7 @@ def get_game_versions():
     """
     print("Fetching game versions")
     data = requests.get(f"{ENDPOINT}/game").json()
+    print(data[0])
```
Infinidoge, line 40:
```suggestion
```

Seems to be a leftover debug line.

Infinidoge (commented):
Something that is up for discussion is how far back the loader versions should be packaged.
Since both Fabric and Quilt are backwards compatible between Minecraft versions, packaging the latest stable version makes sense, and would cut down on a lot of the available packages.

imsofi:
> Something that is up for discussion is how far back the loader versions should be packaged. Since both Fabric and Quilt are backwards compatible between Minecraft versions, packaging the latest stable version makes sense, and would cut down on a lot of the available packages.

One solution to this would be to put the loader version in as an argument that you can override? Instead of doing `fabric-1_19_2-0_10_7` we could have it be something among the lines of `fabric-1_19.overrideAttrs { loader = ... }` as loader versions is rarely important. But it would mean down the line `fabric-1_19_2` will always change its version and not "stay" static. Which can build up old versions that may not work correctly? I am not actually sure how the loaders work for older releases.

I personally have no real problem with keeping the versions as is. As outside of making `nix flake check` rather slow, it doesnt do much harm.

flake.nix
imsofi: Ah i was actually attempting to alphabetically sorting them. But i have seem to have failed my ABCs here :sweat_smile:

README.md
imsofi: I seem to have nuked this part of the README by accident in a rebase. Will re-add it.

flake.nix
imsofi: I do think that `our.lib` is still a lot more descriptive that this is our local library only found in this flake, while i worry for confusion in the naming `flake.lib` as it can sound like it is a generic library for flakes.

I am open for changing it, but I think using `flake.lib` would be a downgrade in readability.

Infinidoge:
> But it would mean down the line `fabric-1_19_2` will always change its version and not "stay" static. 

This is already the case, so I don't really see that as a problem.

> something among the lines of `fabric-1_19.overrideAttrs { loader = ... }`

This is actually a pretty good solution, I didn't really think about `override`/`overrideAttrs` much.
I was originally thinking about something like `lib.makeServer { loader = drv; vanillaServer = drv; }`, which would be a bit much.
This does mean that it's pretty simple to just keep all of the loader versions in the lock files, and just update the stable version every so often.

imsofi:
Also since how similar quilt and fabric's hook into the vanilla binary is, we could also make this `overrideAttrs loader` be put on the vanilla packages, and only have fabric and quilt package its loaders `fabric-loader-version` and `quilt-loader-version` together with defaults to all the versions with the latest loader.

Infinidoge:
Yeah I was also thinking about that at some point too, though to me it seemed a bit odd to include bits about mod loading in vanilla derivations, since they're supposed to be 'just' vanilla.

The main question would be "How do we minimize mod infrastructure in vanilla", in order to keep it loader agnostic

imsofi:
Also current `update.py` seems to not handle spaces inside urls correctly. Leading to missing `sha256` hashes. I am just manually updating the few that get affected currently. But we should look to fix that.

README.md
imsofi: I have now bumped the minimum version to 0.13.0.

It removed close to 100k lines off the fabric locks.json :sweat_smile:

imsofi:
> Also current `update.py` seems to not handle spaces inside urls correctly. Leading to missing `sha256` hashes. I am just manually updating the few that get affected currently. But we should look to fix that.

I have now fixed this issue with the newer commits. `nix-prefetch-url` does not like spaces in its urls :P

imsofi:
Okay so a couple of things i would like to finalize before this is ready for review currently are:

- [ ] Make `our.lib` into `lib.our` to avoid sourcing `our` everywhere.
- [ ] Write up a MIGRATION.md or similar file for how to swap to the new format.
- [ ] Possible `packages.minecraftServers` deprecation message. So that people get a nice message than `missing package`.
- [ ] Look into writing a simple example.

Infinidoge:
> Make our.lib into lib.our to avoid sourcing our everywhere.
Wasn't this how it was before?

imsofi:
It may have been, but i did not figure out how to overlay lib.our in a nice way yet. :sweat_smile:

imsofi:
@Infinidoge Would you be interested in having this rebased and applied onto master? Planning to pick this back up

Infinidoge:
Sure, go right ahead!

imsofi:
So i tried a bit. But this turned quite impossible due to the sheer size of lockfiles in this repo, its begun crashing my editor. So it would likely have to be remade from the ground up. Or try to work up an update to how the lockfile structure is to make them smaller.

Infinidoge:
Since the repo has diverged significantly from when this PR was written, I'm going ahead and closing it. Thankfully, a lot of what was brought up in this PR has been added separately. Thanks for the effort you put into it :)

PR #18, opened 2023-01-05
Option for copying files (instead of symlinking)

Misterio77:
Hello!

First of all, thanks a lot for this awesome module!

I've been trying to get a full minecraft server network going, and it's been going great! I've made a couple of changes for my own use, so here's one of them, in case you're interested.

The ideia is that a lot of configuration files (specially plugins) behave terribly when faced with the read-only fs link destinations. Paper, for example, completely crashes if `config/paper.yml` is read-only.
A solution I've come up with is adding a `files` option that behaves similarly to `symlinks`, but it dereferences the file, making it mutable. That way, the plugin/mod/server can behave as expected, while keeping reproducibility (as the files are copied again each time the service starts).

An existing file with the same name is only backed up if: 1) it's not a symlink 2) it's different to the one being copied (that way a useless `.bak` won't replace a potentially useful one).

I've also went ahead and added a `mkdir` command, so that any parent directories are created (both for files and for symlinks).

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -300,10 +300,10 @@ in
               symlinks = {
                 "eula.txt" = eulaFile;
               } // conf.symlinks;
-              files = {
-                "whitelist.json" = mkIf (conf.whitelist != { }) whitelistFile;
-                "server.properties" = mkIf (conf.serverProperties != { }) serverPropertiesFile;
-              } // conf.files;
+              files =
+                (optionalAttrs (conf.whitelist != { }) { "whitelist.json" = whitelistFile; }) //
+                (optionalAttrs (conf.serverProperties != { }) { "server.properties" = serverPropertiesFile; }) //
+                conf.files;
 
               startScript = pkgs.writeScript "minecraft-start-${name}" ''
                 #!${pkgs.runtimeShell}
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -81,6 +81,17 @@ in
       '';
     };
 
+    environmentFile = mkOpt' (types.nullOr types.path) null ''
+      File consisting of lines in the form varname=value to define environment
+      variables for the minecraft servers.
+
+      Secrets (database passwords, secret keys, etc.) can be provided to server
+      files without adding them to the Nix store by defining them in the
+      environment file and referring to them in option
+      <option>services.minecraft-servers.servers.<name>.files</option> with the
+      syntax @varname@.
+    '';
+
     servers = mkOption {
       default = { };
       description = ''
@@ -330,6 +341,8 @@ in
                   GuessMainPID = true;
                   RuntimeDirectory = "minecraft";
                   RuntimeDirectoryPreserve = "yes";
+                  EnvironmentFile = mkIf (cfg.environmentFile != null)
+                    (toString cfg.environmentFile);
                 };
 
                 preStart =
@@ -362,7 +375,11 @@ in
                               mv "${n}" "${n}.bak"
                             fi
                             mkdir -p $(dirname ${n})
-                            cp -L --no-preserve all ${v} ${n}
+                            ${pkgs.gawk}/bin/awk '{
+                              for(varname in ENVIRON)
+                                gsub("@"varname"@", ENVIRON[varname])
+                              print
+                            }' "${v}" > "${n}"
                           '')
                           files));
                   in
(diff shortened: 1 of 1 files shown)


Infinidoge:
Thanks for the contribution!
I'm glad the module is working well for you, and I'll certainly look out for what you do with it in the future, if it ends up being public.
(Note to self: Perhaps an 'Examples in the Wild' section of the readme?)

The addition definitely makes sense, though I do shake my fist in the direction of the plugins and servers that don't gracefully handle read-only files.
(Aside: Did you know that LuckPerms requires write access to your mods/plugins folder to do *its own dependency management?* It really did not like being in the Nix store lol)

The file backup scheme is sound, and is good to see.

One potential problem with this is files that are left over from files being renamed or removed from the Nix-side, since the file under the old name isn't removed.
This is also a problem for `symlinks` that I haven't really figured out yet, and honestly, I'm not sure what a satisfactory solution for it would be. 
This hasn't been much of a problem in my case for a `mods` directory because I use `linkFarmFromDrvs` to make a single folder to symlink in, instead of single-mod symlinks, but the latter use-case should be supported properly, which it isn't yet.
Perhaps a post-exit script to remove files and clean up symlinks? Not sure what the correct solution is.

Overall, looks good to me! I am curious if you have any ideas on how to handle the renamed/moved file problem, though.

Misterio77:
Thanks a lot for your kind words!

> I'm glad the module is working well for you, and I'll certainly look out for what you do with it in the future, if it ends up being public.
(Note to self: Perhaps an 'Examples in the Wild' section of the readme?)

Feel free to take a look (very much WIP!): https://github.com/Misterio77/nix-config/blob/main/hosts/electra/services/minecraft/default.nix

> The addition definitely makes sense, though I do shake my fist in the direction of the plugins and servers that don't gracefully handle read-only files.
(Aside: Did you know that LuckPerms requires write access to your mods/plugins folder to do its own dependency management? It really did not like being in the Nix store lol)

Absolutely insane. There's a few plugins that will refuse to work if they can't "fill in the blanks" in your config, some will refuse to work even if the config is complete but readonly, while a few others simply won't work if there's missing fields at all! 

> This hasn't been much of a problem in my case for a mods directory because I use linkFarmFromDrvs to make a single folder to symlink in, instead of single-mod symlinks, but the latter use-case should be supported properly, which it isn't yet.

I wish plugins did the same as mods and place their configs in a separate directory so I could use `linkFarmFromDrvs`, oh well

> Perhaps a post-exit script to remove files and clean up symlinks? Not sure what the correct solution is.

This is actually something I just tackled right now! Here's the commit from the [`post-stop`](https://github.com/Misterio77/nix-minecraft/tree/post-stop) branch: https://github.com/Misterio77/nix-minecraft/commit/1d38d8a9a16ee3c15bc321664255c4a16c16702c.

I also noticed that `eula.txt`, `server.properties`, and `ops.json` could simply be `files` and `symlinks`, so I grouped them together in that commit.

Infinidoge:
I commented on it in the actual commit self (not sure if you get notifications for that).
My only concern with how it was implemented was the changes to the unlink/remove/backup&move logic.

https://github.com/Misterio77/nix-minecraft/commit/1d38d8a9a16ee3c15bc321664255c4a16c16702c#r95111294

(If I want to get nitpicky, I think the eula/whitelist/server properties actual derivations should be made separately in the let-block, since how they are defined right now feels dense and hard to read, but that doesn't really matter)

Misterio77:
Oops didn't quite see your previous comment. I'm gonna go ahead and pull the commit into this branch and integrate your feedback!

Misterio77:
Rebased and ready to go!

Misterio77:
I've recently made a change that allows users to specify placeholders on files that will be replaced by values from an `environmentFile` upon copying them. This allows servers to safely configure plugins with databases credentials and other secrets. 

I'll push the change to this branch, but I can open a separate PR if you prefer.

Infinidoge:
Hmm,
I'd say leave it in the PR for now, since it's somewhat intrinsically tied to the `files` option.

Misterio77:
Hey, I've been away for a little while busy with life stuff, but now everything's awesome and I can get back into this.

Is there anything I can do to help push these features forward? Do you happen to have any new suggestions?

Misterio77:
I'm very interested in this project and I'd love to help make it the absolute best solution for building declarative minecraft networks!

Misterio77:
(Fixed conflicts)

hxr404 (approved):

Infinidoge:
Sorry for taking so long to finally address this. My school semester just ended, so now going back through and getting things done.

PR #20, opened 2023-01-06
module: getExe instead of /bin/minecraft-server

Misterio77:
Use `lib.getExe` instead of assuming the executable is named `minecraft-server`.


Infinidoge:
I have no idea why I didn't know about this before, good catch!

PR #21, opened 2023-01-07
feat: add paper server packages

Misterio77:
Package all https://papermc.io servers.

### DIFF
--- a/pkgs/paper-servers/default.nix
+++ b/pkgs/paper-servers/default.nix
@@ -1,0 +1,40 @@
+{ callPackage
+, lib
+, javaPackages
+}:
+let
+  inherit (lib.our) escapeVersion;
+  inherit (lib) nameValuePair flatten last versionOlder mapAttrsToList;
+  versions = lib.importJSON ./lock.json;
+
+  # Remove -build... suffix
+  stripBuild = v: builtins.head (builtins.match "(.*)-build.*" v);
+  # Sort by attribute 'attr' using 'f' function
+  sortBy = attr: f: builtins.sort (a: b: f a.${attr} b.${attr});
+
+  getJavaVersion = v: (builtins.getAttr "openjdk${toString v}" javaPackages.compiler);
+  # https://docs.papermc.io/paper/getting-started#requirements
+  getRecommendedJavaVersion = v:
+    if versionOlder v "1.11.2" then getJavaVersion 8
+    else if versionOlder v "1.16.5" then getJavaVersion 11
+    else if versionOlder v "1.17.1" then getJavaVersion 16
+    else getJavaVersion 17;
+
+  packages = mapAttrsToList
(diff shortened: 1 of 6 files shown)


pkgs/paper-servers/update.py
```diff
@@ -0,0 +1,71 @@
...
+        super().__init__(*args, **kwargs)
+
+    def send(self, request, **kwargs):
+        timeout = kwargs.get("timeout")
+        if timeout is None:
+            kwargs["timeout"] = self.timeout
+        return super().send(request, **kwargs)
+
+
+def make_client():
```
Misterio77, line 29: I had to add some timeout and retries, as it seems the papermc api is a bit unstable (at least on my machine) and often hangs.

pkgs/paper-servers/update.py
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+    json.dump(output, lock, indent=2)
+    lock.write("\n")
+
+
+if __name__ == "__main__":
+    folder = Path(__file__).parent
+    lock_path = folder / "lock.json"
+    main(open(lock_path, "w"), make_client())
```
Misterio77, line 71: I decided to write the lock from scratch every time, as it takes only about 12 seconds to do so (there's much fewer builds when compared to fabric).

pkgs/paper-servers/default.nix
```diff
@@ -0,0 +1,40 @@
...
+      (buildNumber: value: callPackage ./derivation.nix {
+        inherit (value) url sha256;
+        version = "${mcVersion}-build.${buildNumber}";
+        jre_headless = getRecommendedJavaVersion mcVersion;
+      })
+      builds))
+    versions;
+
+  # Latest build for each MC version
+  latestBuilds = sortBy "version" versionOlder (map last packages);
```
Misterio77, line 34: My approach for getting the latest build for each MC version was making a nested list and grabbing the first of each. `packages` looks something like this (each element being a derivation): 
```
[ [ paper-1.17.1-build.1 paper-1.17.1-build.2 ... ] [ paper-1.18-build.1 paper-1.18-build.2 ... ] ... ]
```

pkgs/paper-servers/update.py
Infinidoge: Darn you unstable APIs!
That being said, definitely a good idea to have.

pkgs/paper-servers/update.py
Infinidoge: Makes sense, Fabric/Quilt definitely have a lot more builds with the sheer number of loader version/minecraft version combinations.

Infinidoge (changes requested):
Looks good overall!

That being said, the README needs to be updated to reflect the changes before this can be merged.

Once that's done, I'll smoke test before merging.

PR #22, opened 2023-01-07
add github action for auto updates

Misterio77:
Automate running the `update.py` scripts and committing the results. An example of a commit it creates: https://github.com/Misterio77/nix-minecraft/commit/b97a92e39228917aa19620a1897e8e3958be83b1

### DIFF
--- a/.github/workflows/update-locks.yml
+++ b/.github/workflows/update-locks.yml
@@ -1,0 +1,24 @@
+name: Update package locks
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


Infinidoge:
I've been meaning to do this, I just didn't know enough about GHA to do it properly.
As a nitpick, could you include an if statement to check if `update.py` exists before trying to run it?
(It is possible that some packages may be added without an auto-update script, and I'd prefer it check rather than break.)

PR #24, opened 2023-01-14
feat: package velocity

Misterio77:
Here's velocity proxy packaged as well.


flake.nix
```diff
@@ -26,12 +26,20 @@
...
           paperServers = callPackage ./pkgs/paper-servers { };
-          minecraftServers = vanillaServers // fabricServers // quiltServers // legacyFabricServers // paperServers;
+          velocityServers = callPackage ./pkgs/velocity-servers { };
+          minecraftServers =
+            vanillaServers
+            // fabricServers
+            // quiltServers
+            // legacyFabricServers
+            // paperServers
+            // velocityServers;
```
Misterio77, line 36: Should servers such as velocity and bungeecord be included in `minecraftVersions`?

Infinidoge: Hmm, depends on if they can be dropped into `services.minecraft-server` from Nixpkgs.
The priority is to make sure compatibility is kept with that.

Since I don't think velocity, bungeecord, etc, can be dropped in like that, I'd say they shouldn't be included there.

README.md
```diff
@@ -89,6 +89,14 @@ Since Legacy Fabric does not have a defined newest version to target, it lacks a
 
 `paperServers` functions the same as `fabricServers`, but with the Paper server software.
 
+### `velocityServers.*`
+
+[Source](./pkgs/paper-servers)
```
Infinidoge, line 94:
```suggestion
[Source](./pkgs/velocity-servers)
```

Infinidoge (changes requested):
Outside of this, and the discussion about `minecraftServers`, this looks pretty good to me!

Infinidoge (changes requested):
(Sorry for the delay in getting back to this, have been a bit busy this past week.)

I think the only problem left is that `velocityServers.velocity`/`velocity-server` can point to snapshot versions, when it should point to the latest stable version.
Once that is fixed, I'll go ahead and merge.

Infinidoge:
@Misterio77 Any chance the above could be fixed so this can be merged?

Misterio77:
Sorry for the delay!

> `velocityServers.velocity`/`velocity-server` can point to snapshot versions

Okay so, hm, the file names are a little misleading. The latest non-snapshot build was released in 2021(!), and is really out of date. I think they're simply calling everything snapshot as it's not a "fully duper super stable" software yet.

What we do have are experimental builds (https://papermc.io/downloads#Velocity marked in red here), so everything that's not marked as experimental is considered "stable". I can check for this in the script and avoid marking them as the latest stable version, give me a sec.

Infinidoge (approved):
Looks good to me, thanks!

PR #25, opened 2023-01-14
fix firewall inconsistencies

Misterio77:
There were a few errors with the `openFirewall` implementation:

- Server port was UDP instead of TCP
- Query port was TCP instead of UDP
- UDP ports were opened on TCP needlessly (this hid the above errors)

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -232,19 +232,17 @@ in
         networking.firewall =
           let
             toOpen = filterAttrs (_: cfg: cfg.openFirewall) servers;
-            UDPPorts = mapAttrsToList (name: conf: conf.serverProperties.server-port or 25565) toOpen;
-            TCPPorts = concatLists
-              (mapAttrsToList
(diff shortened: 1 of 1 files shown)


Infinidoge:
Huh, Query is UDP and not TCP?
This needs to be fixed in base Nixpkgs then, since I pretty much lifted that entire setup straight from Nixpkgs, modifying it to account for multiple servers.

Misterio77:
> Huh, Query is UDP and not TCP?

Yup!

> This needs to be fixed in base Nixpkgs then, since I pretty much lifted that entire setup straight from Nixpkgs, modifying it to account for multiple servers. ([For reference](https://github.com/Infinidoge/nixpkgs/blob/e40de94c4266c3c21f6c7e8f2e85c9fef25078d2/nixos/modules/services/games/minecraft-server.nix#L266-L274))

Hmmm we should fix that

Infinidoge:
Sorry for taking so long, thanks for the contribution!

PR #29, opened 2023-03-12
Write file and symlinks from expressions

Misterio77:
Building upon https://github.com/Infinidoge/nix-minecraft/pull/18.

Diff from #18 to this:  https://github.com/Infinidoge/nix-minecraft/compare/Misterio77:deref-symlinks...Misterio77:file-formats

So, basically file/symlink items can now either be:

- The usual string-like (string, path, derivation) with the same behaviour we have today.
- A submodule that contains a `value` (that can be any sort of list/attrset) and a `format` (such as `pkgs.formats.json { }`).
  - The `format` may be null, in this case it will be infered (if possible) from the filename.

This is really really convenient for writing configuration files, as you can simply do them in nixlang without having to call any sort of function to write it.

Here's a real-world example:
```nix
{ pkgs, ... }: {
  services.minecraft-servers.servers.velocity = {
    enable = true;
    # ...
    files = {
      "velocity.toml".value = {
        config-version = "2.5";
        bind = "0.0.0.0:25565";
        player-info-forwarding-mode = "modern";
        forwarding-secret-file = "";
        forwarding-secret = "@VELOCITY_FORWARDING_SECRET@";
        online-mode = true;
        servers = {
          survival = "localhost:25561";
          limbo = "localhost:25560";
          try = [ "survival" "limbo" ];
        };
        forced-hosts = { };
        query = {
          enabled = true;
          port = 25565;
        };
        # ...
      };
      "plugins/librelogin/config.conf".value = {
        allowed-commands-while-unauthorized = [
          "login"
          "register"
          "2fa"
          "2faconfirm"
        ];
        auto-register = true;
        # ...
      };
      # We don't know what format .conf is, so specify it.
      "plugins/librelogin/config.conf".format = pkgs.formats.json { };
    };
  };
}
```

A nice consequence is that `whitelist.json`, `server.properties`, and `eula.txt` also don't have to be manually written with `writeText`.

**TODO**: document this in `files` and `symlinks` descriptions.

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -21,6 +21,57 @@ let
     example = true;
   };
 
+  normalizeFiles = files: mapAttrs configToPath (filterAttrs (_: nonEmptyValue) files);
+  nonEmptyValue = x: nonEmpty x && (x ? value -> nonEmpty x.value);
+  nonEmpty = x: x != { } && x != [ ];
+
+  configToPath = name: config:
+    if isStringLike config # Includes paths and packages
+    then config
+    else (getFormat name config).generate name config.value;
+  getFormat = name: config:
+    if config ? format && config.format != null
+    then config.format
+    else inferFormat name;
+  inferFormat = name:
+    let
+      error = throw "nix-minecraft: Could not infer format from file '${name}'. Specify one using 'format'.";
+      extension = builtins.match "[^.]*\\.(.+)" name;
+    in
+    if extension != null && extension != [ ]
+    then formatExtensions.${head extension} or error
+    else error;
+
+  formatExtensions = with pkgs.formats; {
+    "yml" = yaml { };
+    "yaml" = yaml { };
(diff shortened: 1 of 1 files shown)


Infinidoge (commented):
Overall, definitely agree that it is very useful!

Something that is worth considering is automatically including the "${filename} is managed by NixOS configuration" comment at the top of the file based on the format, since it seems like that is lost in this change.

Though i will admit that I'm not sure how that would be accomplished.

Misterio77:
This should help us with adding comments: https://github.com/NixOS/nixpkgs/pull/224456

Misterio77:
Hey @Infinidoge, I think it'll still take a while to merge the comments PR, do you think we can perhaps merge this one as-is? I can work on the comments later.

Infinidoge:
Once again sorry for this taking so long! Thank you very much for the contribution, and for your patience with me!

PR #30, opened 2023-04-05
fix: un-break nix flake show by moving packages into legacyPackages

hxr404:
The legacyPackages output was created for nixpkgs which has it's derivations inside of nested attribute sets instead of a single top-level set.
Since this flake also uses nested sets for its packages we can simply make use of this to unbreak it.

Temporary fix until #11 is done
(or welp perhaps even a permanent fix in case #11 doesn't get picked up again)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Documentation of major changes, newest first.
 
+## 2023-04-05: Deprecation of `packages` flake output
+Since the packages are currently wrapped inside of attrsets, the flake is more or less "broken". According to the specs, the `packages` outputs may only contain derivations, no sets.
+Nixpkgs doesn't follow that structure either, that's why it uses the `legacyPackages` output instead. Doing the same thing with this flake "un-breaks" things like `nix flake show`
+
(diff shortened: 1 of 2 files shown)


Infinidoge:
Hmm.The issue is that this is a breaking change for anyone not using the overlay.

Infinidoge:
A deprecation notice would be good enough for me.

I might take the opportunity to do a little refactoring to put some packages in the top level without the attrsets.

hxr404:
Ok cool, thx, will do that in a sec!

(kinda) unrelated question: Would you be interested in upstreaming this?
If yes, perhaps simply switching to legacyPackages altogether would probably be the best idea anyway.

Btw, do you have a matrix account or some similar way to contact you?
I'd like to contribute a few PRs (if you don't mind) so it would probably make sense to coordinate that a bit to save your (and my) time.
Alternatively I could just post here on GitHub directly, but I think I'd better ask.

Infinidoge:
I have most of what would be required for this PR locally already, just thinking of how best I should handle pulling in the extra packages.

I have a Discord account at `Infinidoge#1337`, which is my preferred method of communication.
I also have a Matrix account `@infinidoge:matrix.org`, however I make no promises about active responses there, as I do not check it regularly.

PR #31, opened 2023-04-05
fix: replace stdenv with stdenvNoCC to remove gcc dependency

hxr404:
Now that I think of it it's actually weird for nixpkgs to include a C compiler by default instead of the other way round :thinking: 
Anyway, here's the PR


Infinidoge (approved):
Good find!

PR #33, opened 2023-05-12
feat: remove deprecated outputs.packages attribute

imsofi:
Due to the nature of how the packages are exposed in nix-minecraft currently, they do not fit the pattern as required by flakes. This PR removes the packages attribute so to fully rely on legacyPackages instead.

This is done due to all nix tooling that supports flakes complain about the attributes of nix-minecraft (nil_ls, nix flake, etc).

This also closes #11 simply because this repository has grown too much to be able to upkeep such a huge PR. It will likely get repurposed into a hard fork with a smaller scale, current plan being making it fabric only.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -49,32 +49,5 @@
         inherit system;
         config = { allowUnfree = true; };
       });
-
-      packages = {
-        inherit (legacyPackages)
-          vanilla-server
-          fabric-server
-          quilt-server
-          paper-server
-          velocity-server
-          minecraft-server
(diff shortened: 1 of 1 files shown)


Infinidoge:
Worth noting, the `packages` attribute as a whole isn't deprecated. Rather, the attr sets within it are. (Note the block that maps the warning.)

Also, funnily enough, I actually already have a commit ready for this: https://github.com/Infinidoge/nix-minecraft/commit/1dcc182a45bf641bbf175d7a169049f1b66269f0.

Issue #12, opened 2022-09-01
Packwiz support


oatmealine:
[Packwiz](https://packwiz.infra.link/) support would be greatly appreciated, as manually fetching every mod through `fetchModrinth` isn't the best - I'd really like being able to point it at a `pack.toml` URL and have it download the bootstrapper, run it, etc. all on its own.

imsofi:
I am currently working on [packwiz2nix](https://github.com/imsofi/packwiz2nix) tooling for this exact use-case. Its not yet published but will be soon.

For now you could use the imperative way of adding [packwiz-installer-bootstrap](https://github.com/packwiz/packwiz-installer-bootstrap/) to your minecraft server root (`/var/lib/minecraft/`), then inside your NixOS configuration add a `ExecStartPre` to your systemd service file similar to whats described in its [tutorial](https://packwiz.infra.link/tutorials/installing/packwiz-installer/):

> java -jar packwiz-installer-bootstrap.jar -g -s server https://[your-server]/pack.toml

oatmealine:
As an alternative, [unsup](https://git.sleeping.town/unascribed/unsup) has worked quite well for me and supports the Packwiz format:

```nix
let
  # https://git.sleeping.town/unascribed/unsup/releases
  unsup = pkgs.fetchurl {
    url = "https://git.sleeping.town/attachments/c521d178-8938-40a5-b21b-0333eef4099e";
    sha256 = "c5bd49784392b651e4bc71fe57976f5b4fb14f09e0e23183ae5b94a821ae4756";
  };
  packLocation = "https://example.com/pack.toml";
  unsupIni = ''
    version=1
    preset=minecraft

    source_format=packwiz
    source=${packLocation}

    force_env=server
    no_gui=true

    [flavors]
    shaders=no_shaders
  '';
in {
  # ...
  symlinks = {
    "unsup.ini" = pkgs.writeTextFile {
      name = "unsup.ini";
      text = unsupIni;
    };
  };
  jvmOpts = "-Xmx4G -javaagent:${unsup}"
}
```

Infinidoge:
If `fetchPackwizModpack` isn't sufficient, this issue can be reopened if necessary.

TheAyes:
I would like to add an request here as I think it might be fitting.

In my use case I manage a Server for my friends, I would like to manage the Modpack via packwiz. Since in my case I'm basing my modpack off of an existing modpack I would need a way to either:
[...]

PR #35, opened 2023-05-17
add buildPackwizModpack

Misterio77:
Adds a build-support function: `fetchPackwizModpack`.

It uses FOD (fixed output derivation) to fetch an entire Packwiz pack, given its `pack.toml` URL.

You can then easily use the `symlinks` and `files` options to put mods, configs, and anything else the pack provides, in a granular way.

There's also two cool other features:
- The built modpack exports a `manifest` attribute, that is the `pack.toml` imported as a Nix expr. This lets you, for example, get the MC and Fabric version your server should run.
    - This, by default, uses IFD (for convenience). They user may alternatively specify a `manifestHash`, which will then be used with `builtins.fetchurl` to avoid IFD.
- There's also a `addFiles` attribute that lets you include/overwrite additional files on the modpack so, for example, you can add server-specific mods that the pack does not include.

This fixes https://github.com/Infinidoge/nix-minecraft/issues/12

An alternative is [`packwiz2nix`](https://github.com/getchoo/packwiz2nix). It's a smarter implementation but, as of now, it has some limitations that hurt its usability (curseforge does not work, modpacks must have additional hashes added, etc). In the future, when it's more mature and handles all use cases, this function could use `packwiz2nix`'s implementation instead.


Misterio77:
I've added documentation to it (hope we can improve documentation on other parts of the repo in the future, too!), but here's an additional usage example:

https://github.com/Misterio77/nix-config/blob/2936608/hosts/celaeno/services/minecraft/servers/modpack/default.nix

Infinidoge (changes requested):
The derivation itself can be moved under `./pkgs/tools` with the folder named as `fetchPackwizModpack` and it will be automatically included in the flake by the `rakeLeaves` invocation, and wouldn't be included in `packages` since that is explicit.

I'm still not quite sure on which would be the best way. I personally like file-based hierarchies as opposed to specifying it explicitly, however I think that only really works when everything ends up having a similar structure on into the future, which isn't necessarily the case. Still not quite sure on it.

README.md
```diff
@@ -151,6 +221,7 @@ nix run github:Infinidoge/nix-minecraft#nix-modrinth-prefetch -- versionid
 
 This `fetchurl` invocation directly fetches the mod, and can be copy-pasted to wherever necessary.
 
+
```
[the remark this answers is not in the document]
Misterio77: Nope, added on accident haha

README.md
```diff
@@ -121,6 +122,77 @@ For convenience, `velocityServers.velocity` is equivalent to the latest version.
 
 `vanillaServers // fabricServers // quiltServers // legacyFabricServers // paperServers`. Will be used most often as it contains all of the different server versions across each mod loader. When using the overlay, this will replace the Nixpkgs `minecraftServers`.
 
+### `fetchPackwizModpack`
+
+[Source](./pkgs/build-support/fetch-packwiz-modpack)
```
Infinidoge, line 127:
```suggestion
[Source](./pkgs/tools/fetchPackwizModpack)
```

README.md
```diff
@@ -57,7 +58,7 @@ See [TODO.md](./TODO.md).
 
 ## Packages
 
-All of these are found under `legacyPackages`, since they are attrsets of derivations.
+All of these are found under `legacyPackages`, since they are not simple derivations (i.e. an attrset of derivation, or a function that returns a derivation).
```
Infinidoge, line 61:
```suggestion
All of these are found under `legacyPackages`, since they are not derivations (i.e. an attrset of derivation, or a function that returns a derivation).
```

README.md
```diff
@@ -121,6 +122,77 @@ For convenience, `velocityServers.velocity` is equivalent to the latest version.
...
+**Note**: Using `manifest`, by default, will cause [IFD](https://nixos.wiki/wiki/Import_From_Derivation). If you want to avoid IFD while still having access to `manifest`, simply pass a `manifestHash` to the `fetchPackwizModpack` function, it will then fetch the manifest through `builtins.fetchurl`.
+
+Additionally, you can override/add files (e.g. server-specific mods) on the pack through `addFiles`. For example:
+
+```nix
+let
+  modpack = pkgs.fetchPackwizModpack {
+    url = "https://github.com/Misterio77/Modpack/raw/0.2.9/pack.toml";
+    packHash = "sha256-L5RiSktqtSQBDecVfGj1iDaXV+E90zrNEcf4jtsg+wk=";
+  }.addFiles {
```
Infinidoge, line 185:
```suggestion
  modpack = (pkgs.fetchPackwizModpack {
    url = "https://github.com/Misterio77/Modpack/raw/0.2.9/pack.toml";
    packHash = "sha256-L5RiSktqtSQBDecVfGj1iDaXV+E90zrNEcf4jtsg+wk=";
  }).addFiles {
```
The parenthesis are necessary here, since otherwise it would try to get `addFiles` from the input to `fetchPackwizModpack`.

Infinidoge (changes requested):

Misterio77:
Thanks for the feedback!

Infinidoge (approved):

PR #38, opened 2023-06-09
tests: init

Misterio77:
Hi! This is me again, with the first of 4 PRs this time xD

This PR adds simple NixOS integration tests, which can be run with `nix flake check` (or individually with `nix build .#checks.x86_64-linux.simple`).

For now, there's two:
- Simple: runs the vanilla server and check that it actually started
- Files & Symlinks: runs the paper server, with two nix expression-generated files (ops.json and paper.yml), as well as symlink (a vanilla jar that is nescessary to avoid paper trying to download it at runtime).
    - In a follow-up PR (will open it in a few minutes), I will use this to test command execution:
        - I de-op `Misterio7x`, this mutates the ops.json, so validates that `files` are mutable.
        - I run a unknown command and grep the output, validating that the `paper.yml` `unknown-command` is actually used.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -39,6 +39,17 @@
         } // (
           builtins.mapAttrs (n: v: callPackage v { }) (self.lib.rakeLeaves ./pkgs/tools)
         );
+
+      mkTests = pkgs:
+        let
+          callPackage = pkgs.newScope {
+            inherit (self) outputs;
+            lib = pkgs.lib.extend (_: _: { our = self.lib; });
+          };
+        in
+          if (pkgs.stdenv.isLinux)
+          then builtins.mapAttrs (n: v: callPackage v { }) (self.lib.rakeLeaves ./tests)
+          else {};
     in
     {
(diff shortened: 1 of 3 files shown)


Infinidoge:
Looks pretty good to me. Will do local testing to make sure everything is good before merging.

<Aside>
Darn you Paper! Why must you download things at runtime!

It might be a good idea to put the corresponding vanilla jar for a paper server into the paper derivation's passthru.
Also would be a good idea to document the need for that in the readme.
</Aside>

Misterio77:
For normal usage, we usually can assume the user will have internet connection, but I fully agree, we should totally add that to passthrough and document it; I'll do this in a minute.

Misterio77:
There you go: https://github.com/Infinidoge/nix-minecraft/pull/44

Infinidoge:
Sorry for taking so long to merge this. While I do have the excuse that I've been busy with college, it really shouldn't have taken so long to merge a PR of this size. Thanks again for the contributions!

PR #39, opened 2023-06-09
Harden module, use systemd.tmpfiles to create dirs

Misterio77:
This PR hardens the minecraft service, and in that process uses systemd to create directories (and set permissions), instead of relying on scripts.

Try running `systemd-analyze security`, and you'll see that the current score is < 1. With this PR, it increases to 3.

I also changed the umask from `002` to `007`, this means new files will be `770` and not `775`. I think this is better, as other users (that are not in the group) should not have any access to the files whatsoever.

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -266,6 +266,7 @@ in
             description = "Minecraft server service user";
             home = cfg.dataDir;
             createHome = true;
+            homeMode = "770";
             isSystemUser = true;
             group = "minecraft";
           };
@@ -315,7 +316,7 @@ in
           };
 
         systemd.tmpfiles.rules = mapAttrsToList (name: _:
-          "d '${cfg.dataDir}/${name}' 0700 ${cfg.user} - - -"
+          "d '${cfg.dataDir}/${name}' 0770 ${cfg.user} - - -"
         ) servers;
 
         systemd.services = mapAttrs'
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -265,6 +265,7 @@ in
           users.minecraft = {
             description = "Minecraft server service user";
             home = cfg.dataDir;
+            createHome = true;
             isSystemUser = true;
             group = "minecraft";
           };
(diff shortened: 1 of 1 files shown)


Misterio77:
You can check this commit passes the test by cherry-picking the `tests` branch:

```bash
git remote add misterio git@github.com:misterio77/nix-minecraft
git cherry-pick -n misterio/tests

nix flake check

git restore
git remote rm misterio
```

Infinidoge:
TIL `systemd-analyze security` exists. Sometimes I wonder what commands *aren't* in systemd.

modules/minecraft-servers.nix
```diff
@@ -313,16 +314,13 @@ in
             allowedTCPPorts = flatten (mapAttrsToList getTCPPorts toOpen);
           };
 
-        system.activationScripts.minecraft-server-data-dir.text = ''
-          mkdir -p ${cfg.dataDir}
-          chown minecraft:minecraft ${cfg.dataDir}
-          chmod -R 775 ${cfg.dataDir}
-        '';
+        systemd.tmpfiles.rules = mapAttrsToList (name: _:
+          "d '${cfg.dataDir}/${name}' 0700 ${cfg.user} - - -"
```
Infinidoge, line 318: Doesn't that permission octal mean that only the `minecraft` user has access to the data directory?
I believe the reason why I didn't use `createHome` in the first place was because that setup made it so only the `minecraft` user could access the server files, when everybody in the group should be able to access the files.

Infinidoge (commented):
I don't know enough about the internals of systemd to know what all of the hardening options mean, so I'll take your word on them. I do have some concern about the creation and mode of the data directory, though.

I really need to look more into systemd tmpfiles though, it's much more concise.

modules/minecraft-servers.nix
Misterio77: Oops, that was an oversight on my part, it was supposed to be `770`.

Misterio77: About `createHome`, you can use `homeMode` to set the permissions (it defaults to `700`)!

Misterio77: Should be fixed by https://github.com/Infinidoge/nix-minecraft/pull/39/commits/b2256f476cdbf1e4fce46ff488533b94a467011e

PR #42, opened 2023-06-10
module: fix command filename quoting

Misterio77:
The commands would fail when the file names contain a `'`


Infinidoge:
Name a more pervasive problem than shell quoting.

PR #43, opened 2023-06-10
lib: add collectFiles and collectFilesAt

Misterio77:
This is a lib function that recursively reads a directory, and returns them in the format expected by `files`/`symlinks`. This makes it possible to, for example, link all mods from a modpack, and still include additional mods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,13 +146,56 @@ in
     symlinks = {
       "mods" = "${modpack}/mods";
     };
+    files = {
+      "config" = "${modpack}/config";
+      "config/mod1.yml" = "${modpack}/config/mod1.yml";
+      "config/mod2.conf" = "${modpack}/config/mod2.conf";
+      # You can add files not on the modpack, of course
+      "config/server-specific.conf".value = {
+        example = "foo-bar";
+      };
+    };
   };
 }
 ```
 
-This will symlink the modpack's final `mods` directory into the server's `mods` directory. You can also do this for `config`, or any files in the modpack you're interested in, in a granular way.
+This will symlink the modpack's final `mods` directory into the server's `mods` directory, and copy the specified config files into `config`. You can also do this for any files in the modpack you're interested in, in a granular way.
 
 **Note**: Be sure to use a stable URL (e.g. a git tag/commit) to the manifest, as it changing will cause the derivation to generate a different hash, breaking the build until you change it.
+
+Adding config files one by-one is boring. Trying to include an additional mod will also fail, as `mods` is read-only and symlinked as is. To solve these issues, this flake provides a `collectFilesAt` function, that recursively collects every file on the given directory (e.g. `config`), and returns an attribute set in the format `symlinks`/`files` expects. For example:
+
+```nix
+{ inputs, ... }:
+let
+  inherit (inputs.nix-minecraft.lib) collectFilesAt;
+  modpack = pkgs.fetchPackwizModpack {
+    url = "https://github.com/Misterio77/Modpack/raw/0.2.9/pack.toml";
+    packHash = "sha256-L5RiSktqtSQBDecVfGj1iDaXV+E90zrNEcf4jtsg+wk=";
+  };
+in
+{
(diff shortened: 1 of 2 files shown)


Misterio77:
An example on how it works:

```
nix-repl> lib.collectFilesAt modpack "config"
{
  "config/appliedenergistics2/common.json" = "/nix/store/ihsw8ifdmv7xjd3v1c7zqbfbbqkqvzar-packwiz-pack-0.2.18//config/appliedenergistics2/common.json";
  "config/bclib/main.json" = "/nix/store/ihsw8ifdmv7xjd3v1c7zqbfbbqkqvzar-packwiz-pack-0.2.18//config/bclib/main.json";
  "config/bclib/server.json" = "/nix/store/ihsw8ifdmv7xjd3v1c7zqbfbbqkqvzar-packwiz-pack-0.2.18//config/bclib/server.json";
  "config/byg/biomepedia.json5" = "/nix/store/ihsw8ifdmv7xjd3v1c7zqbfbbqkqvzar-packwiz-pack-0.2.18//config/byg/biomepedia.json5";
  "config/byg/client/biomepedia_inventory.json5" = "/nix/store/ihsw8ifdmv7xjd3v1c7zqbfbbqkqvzar-packwiz-pack-0.2.18//config/byg/client/biomepedia_inventory.json5";
  "config/charm.toml" = "/nix/store/ihsw8ifdmv7xjd3v1c7zqbfbbqkqvzar-packwiz-pack-0.2.18//config/charm.toml";
    # ...
}
```

Infinidoge:
Would it be good to have `collectFiles <path>`, with `collectFilesAt <path> <subdir>` being something like (in pseudo-code) `addPrefixToAttrKeys subdir (collectFiles "${path}/${subdir}")`?

If `files` / `symlinks` let you do something like
```nix
files = {
  config = {
    "file1" = ...;
    "file2" = ...;
  };
};
```
to create
```
config/file1
config/file2
```
Then `collectFiles` would be quite useful, but as it stands the system doesn't do that that, so I'm not really sure what cases where you'd need `collectFiles` on its own, unless you want to do `addPrefix "config" (collectFiles ./config)` instead of `collectFilesAt . "config"`

Misterio77:
Hey! Sorry for taking so long to respond.

> Would it be good to have collectFiles <path>, with collectFilesAt <path> <subdir> being something like (in pseudo-code) addPrefixToAttrKeys subdir (collectFiles "${path}/${subdir}")?

This is a very good idea and looks cleaner, will def get that in.

> If files / symlinks let you do something like

This sounds super nice and actually kinda solves some concerns with composability, I think.

--- 

Sadly my current design for this PR has a major footgun.

We currently cleanup files when shutting down the server, but removing files/symlinks entries from your nixos config will NOT remove the files (as the service isn't stopped with the previous generation's stop script)! This isn't a major concern when using a single derivation for a big-ish directory, for example:
```nix
  services.minecraft-servers.servers.cool-modpack = {
    enable = true;
    package = pkgs.fabricServers.fabric-1_18_2-0_14_9;
    symlinks =  "${modpack}/mods"
};
```
Removing mods from `modpack` will propagate these changes to your server. The server might not cleanup the `mods`, but this isn't a problem as an empty directory does nothing on its own.

Once the entry is a leaf, that is, a single file, things get messy. As mentioned, the stop script will remove what's _currently_ defined, not what was _previously_ there. We'll have to special case this somehow. A few solutions I thought of:
- Keep track of managed files. Maybe a hidden file on the server root dir?
- Clean up more agressively. Always delete symlinks to nix store, or try and detect what is being managed some other way.

Misterio77:
SMH, for every common problem, there is already an elegant solution haha, had no idea about that. Will look into it.

Misterio77:
I think stopIfChanged is pretty limited (e.g. it won't work if we prefer to reload instead), so I decided to implement a file-attrs based solution (#116). With that merged, I plan on finishing this one.

Misterio77:
Should be good to go.

Infinidoge (approved):

PR #47, opened 2023-10-24
Implement Velocity Proxy server module

scottbot95:
Implement Velocity Proxy server module. 

It's basically just a special-purpose version of the minecraft-servers module. We probably could just use the minecraft-servers module but some properties (specifically `serverProperties` and `whitelist`) don't really make any sense for Velocity.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -50,13 +50,16 @@
           };
         in
         optionalAttrs isLinux (mapAttrs (n: v: callPackage v { }) (self.lib.rakeLeaves ./tests));
+      nixosModules = self.lib.rakeLeaves ./modules;
     in
     {
       lib = import ./lib { lib = flake-utils.lib // nixpkgs.lib; };
 
       overlay = final: prev: mkPackages prev;
       overlays.default = self.overlay;
-      nixosModules = self.lib.rakeLeaves ./modules;
+      nixosModules = nixosModules // {
+        default = { imports = builtins.attrValues nixosModules; };
+      };
     } // flake-utils.lib.eachDefaultSystem (system:
     let
       pkgs = import nixpkgs {
--- a/modules/velocity.nix
+++ b/modules/velocity.nix
@@ -1,0 +1,316 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let
+  cfg = config.services.minecraft-servers.velocity;
+  config-version = "2.6";
+
+  mkBoolOpt' = default: description: mkOption {
+    inherit default;
+    description = mdDoc description;
+    type = types.bool;
+    example = true;
+  };
+in
+{
+  options.services.minecraft-servers.velocity = {
+    enable = mkEnableOption "Velocity Minecraft Proxy Server";
+
+    autoStart = mkBoolOpt' true ''
+      Whether to start Velocity on boot.
+      If set to <literal>false</literal>, can still be started with
+      <literal>systemctl start velocity</literal>.
+    '';
+
+    datadir = mkOption {
+      type = types.str;
+      default = "/var/lib/velocity";
+      description = mdDoc ''
+        Directory to store the Velocity server.
+      '';
+    };
+
+    openFirewall = mkBoolOpt' false "Whether to open the velocity listen/query ports";
+
+    package = mkOption {
+      description = "The Velocity proxy package to use.";
+      type = types.package;
+      default = pkgs.velocity-server;
+      defaultText = literalExpression "pkgs.velocity-server";
+    };
+
+    address = mkOption {
+      type = with types; nullOr str;
+      default = null;
+      description = mdDoc ''
+        This tells the proxy to accept connections on a specific IP.
+        By default, Velocity will listen for connections on all IP
+        addresses on the computer on port 25577.
+      '';
+    };
+    port = mkOption {
+      type = types.port;
+      default = 25577;
+      description = mdDoc ''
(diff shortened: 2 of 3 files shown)


flake.nix
```diff
@@ -39,13 +39,16 @@
...
     in
     {
       lib = import ./lib { lib = flake-utils.lib // nixpkgs.lib; };
 
       overlay = final: prev: mkPackages prev;
       overlays.default = self.overlay;
-      nixosModules = self.lib.rakeLeaves ./modules;
+      nixosModules = nixosModules // {
+        default = { imports = builtins.attrValues nixosModules; };
+      };
```
Infinidoge, line 51: Good call, I didn't think about making default import everything.

modules/velocity.nix
```diff
@@ -0,0 +1,304 @@
...
+      default = 25577;
+      description = mdDoc ''
+        Port on which to the proxy will listen for connections
+
+        This only has an affect if <option>services.minecraft-servers.velocity.address</option>
+        is non-null.
+      '';
+    };
+
+    config = {
```
Infinidoge, line 61: I assume these options use `dashed-case` as opposed to `camelCase` to match the velocity config?

modules/velocity.nix
```diff
@@ -0,0 +1,304 @@
...
+      servers = mkOption {
+        type = with types; nullOr (attrsOf str);
+        default = null;
+        description = mdDoc ''
+          Attribute set of server names to addresses. These addresses can be used later in
+          <option>services.minecraft-servers.velocity.config.servers.try</option> or
+          <option>services.minecraft-servers.velocity.config.forced-hosts</option>
+
+          If null, auto-generate server list based off 
+          <option>services.minecraft-servers.servers</option>
```
Infinidoge, line 139: If it is possible, documenting a way to both auto-generate and specify would be nice.

Infinidoge (changes requested):
Overall looks good! I assume this has been tested in your setup and works well?

The main thing I would like before merging this would be to have it run through `nixpkgs-fmt`, as that is the main formatter I use. (I really should add more documentation to the repo for stuff like that...)

modules/velocity.nix
scottbot95: Hmmm interesting idea. I could make the default be an auto-generated attribute set instead of `null`. The downside to that approach is I'm pretty sure that means users wouldn't be able remove the auto-generated servers (although perhaps `mkForce` would work?)

modules/velocity.nix
scottbot95: Correct. The config here is just the config values from Velocity mapped to nix types.

flake.nix
scottbot95: Yeah it's a somewhat common pattern when you start having flakes with multiple modules exported. I added this cuz I was too lazy to have multiple imports in my consuming flake lol.

scottbot95:
I can run the formatter. Might take me a day or two before I can get back to this. I've also seen other flakes that include a style check in the actual flake checks so you can setup some GH actions to run `flake check` and it will fail if the files aren't properly formatted. I will take a look at how that's implemented. If it's easy enough I might just include it in this PR as well (or a separate PR).

Misterio77:
Hi! I have some thoughts on this:

- This can and should be implemented through `minecraft-server.servers.*` instead of `systemd.services.*` directly. The rule of thumb is: always use the higher-level abstraction you can get away with using.
  - I'm pretty sure `velocity.toml` **has** to be mutable. Or it'll blow up in the slightest change or missing key.
  - You're missing on super important features such as `files` and `symlinks`, which you can, by the way, use to implement `config` instead of manually linking `configFile`.
  - Due to that, there's a huge amount of duplicated code. This means improvements to `minecraft-server.servers` are not available here, and increases maintenance burden.
-  Mapping all options like that is a major footgun for maintenance. I recommend using [the `settings` pattern](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md) instead. 
    - What if options change between versions? They do change a lot. Will we have to actively hunt for velocity changelogs and add dozens of version checks and warnings for deprecation? This quickly gets out of hand.
    - As an example, you're missing `player-info-forwarding-mode`.

I hate to be this person but, I don't think we should add specialized modules. Minecraft network setups vary a lot (what if they want two proxies? etc). We should focus effort in improving the base module while always keeping module complexity in check. Most of this is already possible (the convenience is very, very small):
```nix
services.minecraft-servers.servers.velocity = {
  enable = true;
  package = pkgs.velocity-server;
  files."velocity.toml".value = {
    bind = "0.0.0.0:25565";
    forwarding-secret-file = ...;
    servers = (mapAttrs (_: v: "localhost:${toString v.serverProperties.server-port}") config.minecraft-servers.servers) // {
      try = [ "foo" ];
    };
  };
};
```
VS
```nix
services.minecraft-servers.velocity = {
  enable = true;
  address = "0.0.0.0";
  port = 25565;
  forwarding-secret-file = ...;
  config.try = [ "foo" ];
};
```

If we're sure we want a specialized variant, it should be a very, very thin wrapper around `minecraft-servers.servers`.

Infinidoge:
Sorry for not getting to this earlier, university was very busy and draining for me this past year. After looking at this again, I agree with Misterio that this is better left to `minecraft-servers.servers`. The only thing that isn't currently possible with the existing module is automatically generating the secret.

Following from that, I'm closing this PR. Thank you for the contribution, though!

PR #41, opened 2023-06-09
Replace tmux with stdin systemd sockets

Misterio77:
To make reviewing this easier, see this diff: https://github.com/Misterio77/nix-minecraft/compare/60bf3412edbb562da2d253fd5159d985f4e849a1...master

---

This PR, that builds on the previous ones (#38, #39, #40), replaces tmux with a systemd stdin socket. With it, you can easily run commands like so:

```
echo "say Hi" > /run/minecraft/foobar.stdin
```

We now use this to run "stop" (or any custom command) on ExecStop.

To complement `extraReload`, I've added more `extra*` options for different parts of the lifecycle: extra{Post,Pre}{Start,Stop} (preStop and postStart can run console commands, awesome!).

The tests are also expanded, as I mentioned on the testing PR.

PR #52, opened 2023-12-14
module: replace tmux with systemd socket

Misterio77:
Supersedes #41.

---

This PR replaces tmux with a systemd stdin socket. With it, you can easily run commands like so:
```
echo "say Hi" > /run/minecraft/foobar.stdin
```
We now use this to run "stop" (or any custom command) on ExecStop, for a cleaner shutdown.

To complement extraReload, I've added more extra* options for different parts of the lifecycle: extra{Post,Pre}{Start,Stop} (preStop and postStart can run game commands, awesome!).

The tests are also expanded:
- In the simple test: I run `list` and check the output is good.
- In the file/symlinks test:
  - I de-op `Misterio7x` to verify `ops.json` is both mutable and has the correct declaratively-set value
  - I run an unknown command and check the message matches `unknown-command`. As paper mutates `paper.yml` (to fill it up with missing entries) on startup, this verifies that values set through nix-minecraft are kept by it.

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -101,13 +101,6 @@ in
       the server name in this directory, such as <literal>/srv/minecraft/servername</literal>.
     '';
 
-    runDir = mkOpt' types.path "/run/minecraft" ''
-      Directory to place the runtime tmux sockets into.
-      Each server's console will be a tmux socket file in the form of <literal>servername.sock</literal>.
-      To connect to the console, run `tmux -S /run/minecraft/servername.sock attach`,
-      press `Ctrl + b` then `d` to detach.
-    '';
-
     user = mkOption {
       type = types.str;
       default = "minecraft";
@@ -125,8 +118,8 @@ in
       default = "minecraft";
       description = ''
         Name of the group to create and run servers under.
-        In order to modify the server files or attach to the tmux socket,
-        your user must be a part of this group.
+        In order to modify the server files your user must be a part of this
+        group.
         It is recommended to leave this as the default, as it is
         the same group as <option>services.minecraft-server</option>.
       '';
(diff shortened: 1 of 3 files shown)


Infinidoge:
As per my comment in the last PR, a prerequisite for this kind of change is that it *must* be configurable. It should be possible to select between tmux and socket, and tmux would remain the default for a deprecation period.

Misterio77:
Yep! Will do. Just re-created this PR to get discussion and feedback going

Silverdev2482:
I am not familiar with systemd sockets, but do they offer an interactive mode like Tmux? Although not a feature I depend on, it is a convenience I enjoy having.

Silverdev2482:
I think this is ok to put this here instead of a separate issue, could there be a way to configure the socket to be stored in the Minecraft server's directory? That would me a minor convince to me as I have to type shorter directory names.

Infinidoge:
systemd sockets do *not* have an interactive function, which is a major reason why I am so adamant about there being a toggle for it.

Sockets are by design put in the /run runtime directory, as that is the standard place to put sockets. (Additionally, it being in /run is generally shorter than the /srv equivalent: `/run/minecraft/[servername].sock` vs `/srv/minecraft/[servername]/server.sock` (or similar))

Silverdev2482:
The reason I say I would like to be able to configure them to be there is that I'm so often cd'ed there. Although that rational makes sense.

frantisekhanzlikbl:
Hi, @Misterio77, are you still working on this? I very much like the changes and am interested in using them myself. If you're no longer in working on this, would you mind me picking it up to make the socket configurable? Thanks!

Misterio77:
Hey @frantisekhanzlikbl,

Thanks for the interest, feel free to pick it up!

While I think the socket being configurable is a useful addition, I think the main blocker for this is currently making this a toggle as per @Infinidoge's comment

frantisekhanzlikbl:
Ohh, sorry, I didn't express myself too well. :smile:
By “making the socket configurable” I meant “making the usage of socket vs tmux configurable”. I'm currently awfully short on time, but hopefully, I'll be able to get on this in a few weeks.

Infinidoge:
Superseded by #67

PR #54, opened 2023-12-28
module: fix assertion for clashing `dataDir`s

h7x4:
This assertion triggers when `minecraft-server` is not enabled and the `dataDir`s are equal. I'm assuming this is a bug, and the `!` was meant to negate the whole expression (and also swapping the `!=` for an `==`), but I think the implication arrow makes the intent more clear.


Infinidoge:
Yep that's a bug, and a much better expression, good catch!

PR #56, opened 2024-01-11
Add mainProgram to velocity-server derivation

ATiltedTree:
Removes the warning when using the velocity-server derivation with the nixos module:
```
trace: warning: getExe: Package "velocity-3.3.0-SNAPSHOT-build.316" does not have the meta.mainProgram attribute.
We'll assume that the main program has the same name for now, but this behavior is deprecated, because it leads to surprising errors when the assumption does not hold.
If the package has a main program, please set `meta.mainProgram` in its definition to make this warning go away.
Otherwise, if the package does not have a main program, or if you don't control its definition, use getExe' to specify the name to the program, such as lib.getExe' foo "bar".
```


Infinidoge:
Surprised this hadn't been mentioned yet, good catch!

PR #57, opened 2024-01-21
Add mutable whitelist and server.properties options

matthew-salerno:
As a server owner I want to be able to manage my whitelist imperatively, as it often changes. I allowed specifying null for `whitelist` and `serverProperties` to leave the files mutable when not specified, preventing overwriting with empty files. This is mainly intended for whitelists, but I also implemented it for serverProperties for completeness' sake.
- Whitelist tested and confirmed that changes are not overridden on service restart (when whitelist option is `null`)
- serverProperties I assume works but I haven't tested it, the logic is exactly the same as whitelist.

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -208,12 +208,13 @@ in
                   description = "Minecraft UUID";
                 };
               in
-              types.attrsOf minecraftUUID;
-            default = { };
+              types.nullOr (types.attrsOf minecraftUUID);
+            default = null;
             description = ''
               Whitelisted players, only has an effect when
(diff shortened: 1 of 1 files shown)


Infinidoge:
Heya! This is actually a redundant change. If you don't set a whitelist in the options, then `normalizeFiles` already filters it out, so one isn't generated.

Simply setting `serverProperties.white-list = true;` without specifying a whitelist in Nix will achive what you are looking for.

Infinidoge:
The main thing that is needed is documentation.

matthew-salerno:
Hmm I thought my whitelist was getting deleted before but I made a couple changes at once so maybe it was something else. Thank you for taking the time to clarify!

PR #59, opened 2024-01-29
Transfer documentation from module to README

Silverdev2482:
I think this is mostly good but I'm not sure about the syntax of how I named options, I also wouldn't be suprised as if there are a few spelling errors. It still need proofreading by me too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,36 +236,44 @@ Module for hosting multiple servers at once. All of the following are under this
 
 #### `enable`
 
-If enabled, the servers in `services.minecraft-servers.servers` will be created and started as applicable. The data for the servers will be loaded from and saved to `dataDir`
+If enabled, the servers in `services.minecraft-servers.servers` will be created and started as applicable. 
+The data for the servers will be loaded from and saved to `dataDir`, and any sockets will be put in `runDir`.
 
 #### `eula`
 
-Whether you agree to [Mojang's EULA](https://account.mojang.com/documents/minecraft_eula) This option must be set to true to run Minecraft server.
+Whether you agree to [Mojang's EULA](https://account.mojang.com/documents/minecraft_eula). 
+This option must be set to true to run any Minecraft servers.
 
 #### `openFirewall`
 
-Whether to open ports in the firewall for each server. Sets the default for `servers.<name>.openFirewall`. This will only work if the ports are specified in `servers.<name>.serverProperties` otherwise it will simply use the default ports. Remember to change the ports if you running multiple servers. 
+Whether to open ports in the firewall for each server. Sets the default for `servers.<name>.openFirewall`. 
+This will only work if the ports are specified in `servers.<name>.serverProperties`, otherwise it will use the default ports.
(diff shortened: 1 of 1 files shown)


Silverdev2482:
Will `servers.<name.files` Only place non existing files, or will it overwrite them? I would like to specify that in the README.

Silverdev2482:
Looking at it I believe it will put the files into a .bak when modified, but will overwrite the previous.bak file, if I am wrong please correct me.

README.md
```diff
@@ -278,3 +367,6 @@ This example takes an attrset of the IDs and hashes for Modrinth mods, fetches e
   };
 }
 ```
+
+#### `servers.<name>.files`
+Things to copy into this server's data directory. Similar to symlinks, but these are actual files. Useful for configuration files that don't behave well when read-only.
```
Infinidoge, line 372: `files` also has the option for generating files from data, which would be useful to document. CC @Misterio77 for advice on that.

Infinidoge (changes requested):
Thanks for putting in the work of putting it into the file. I'm not too happy with how it is all laid out, but that is mostly my fault for how it is documented. Something I want to look into for the future would be generating the documentation like how it is done in Nixpkgs et al., as that would be the best solution.

Most of the suggestions are things that can just be committed as-is, let me know if you have any thoughts on them.

Silverdev2482:
All suggestions look good to me, I wasn't too sure on format/syntax so I went to what seems the most similar with the existing readme. Not sure if I feel like documenting the files option, as I don't have the best understanding of the nix language. If @Misterio77 feels like documenting that we should wait, or maybe a make separate issue and pr. Other than that this seems all good to merge.

README.md
```diff
@@ -231,17 +231,127 @@ This `fetchurl` invocation directly fetches the mod, and can be copy-pasted to w
...
+  "rcon.password" = "hunter2";
+}
+```
+
+#### `servers.<name>.package`
+
+The Minecraft server package to use. Example:
+`pkgs.minecraftServers.vanilla-1_18_2`
+
+#### `servers.<name>.package`
```
BatteredBunny, line 346: I'm assuming this is supposed to be ``jvmOpts`` instead?

BatteredBunny (commented):
Noticed a typo

README.md
Infinidoge: Surprised I didn't notice that, good catch.
```suggestion
#### `servers.<name>.jvmOpts`
```

Silverdev2482:
Should that issue be created and this merged? Or do we add that to this PR and wait for it to be documented?

Silverdev2482:
don't mean any disrespect, but should we just merge this now? If I am not mistaken there isn't anything I need to do, and Misterio hasn't responded. Every other suggestion looks fine and has been approved.

Infinidoge:
Sorry for the wait, summer has been pretty busy for me so even though this is small and straight forward, it kept getting pushed back by the other things I've been doing.

PR #61, opened 2024-03-25
module: use user and group from configuration

jackwilsdon:
Previously, `cfg.user` and `cfg.group` were not used by the systemd service. This commit fixes this and also disables creation of the `minecraft` user and group if the user and group are changed.

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -277,7 +277,7 @@ in
       in
       {
         users = {
-          users.minecraft = {
+          users.minecraft = mkIf (cfg.user == "minecraft") {
             description = "Minecraft server service user";
             home = cfg.dataDir;
             createHome = true;
@@ -285,7 +285,7 @@ in
             isSystemUser = true;
             group = "minecraft";
           };
-          groups.minecraft = { };
+          groups.minecraft = mkIf (cfg.group == "minecraft") { };
         };
 
         assertions = [
@@ -387,7 +387,8 @@ in
                   ExecStop = "${stopScript} $MAINPID";
                   Restart = conf.restart;
                   WorkingDirectory = "${cfg.dataDir}/${name}";
-                  User = "minecraft";
+                  User = cfg.user;
+                  Group = cfg.group;
                   Type = "forking";
                   GuessMainPID = true;
                   RuntimeDirectory = "minecraft";


modules/minecraft-servers.nix
```diff
@@ -277,15 +277,15 @@ in
       in
       {
         users = {
-          users.minecraft = {
+          users.minecraft = mkIf (cfg.user == "minecraft") {
```
Infinidoge, line 280: It might be better if this was `users.${cfg.user}` instead of disabling it outright.

modules/minecraft-servers.nix
```diff
@@ -277,15 +277,15 @@ in
...
             description = "Minecraft server service user";
             home = cfg.dataDir;
             createHome = true;
             homeMode = "770";
             isSystemUser = true;
             group = "minecraft";
           };
-          groups.minecraft = { };
+          groups.minecraft = mkIf (cfg.group == "minecraft") { };
```
Infinidoge, line 288: See previous comment, this might be better as `groups.${cfg.group}` instead.

modules/minecraft-servers.nix
```diff
@@ -387,7 +387,8 @@ in
                   ExecStop = "${stopScript} $MAINPID";
                   Restart = conf.restart;
                   WorkingDirectory = "${cfg.dataDir}/${name}";
-                  User = "minecraft";
+                  User = cfg.user;
+                  Group = cfg.user;
```
Infinidoge, line 391:
```suggestion
                  Group = cfg.group;
```

Infinidoge (changes requested):
I am... honestly not sure how I never noticed that fact. Configuring the user/group was definitely an afterthought, since I never needed to do that personally, but it not actually doing the important part of creating that user and using it for the service is a little bit embarrassing.

Good catch!

modules/minecraft-servers.nix
jackwilsdon: Services that allow configuring user/group usually allow the user to create them if they specify a non-default value, e.g.:

https://github.com/NixOS/nixpkgs/blob/3b597532faf630d5053fe1072d36c4de4f773609/nixos/modules/services/misc/forgejo.nix#L634-L645

modules/minecraft-servers.nix
jackwilsdon: :facepalm: fixed

modules/minecraft-servers.nix
Infinidoge: Makes sense. It would probably be a good idea to add an assertion that the group/user exists then, if there is precedent for that in Nixpkgs. (I suppose mutable users makes that a bit messy)

Infinidoge: From what I've seen that is the case, yeah. This should be fine as-is then.

Infinidoge (approved):

PR #62, opened 2024-03-28
module: wait for tmux to exit in stopScript

Silveere:
This stops the server from being immediately force-killed by systemd. Any process not cleaned up by `ExecStop` is killed by systemd immediately after all `ExecStop` scripts exit.

This also changes the process detection, since `$MAINPID` doesn't seem to exist (at least from the perspective of the `ExecStop` script). I'm not sure if this is due to `ProtectProc=invisible`, something with systemd's PID detection, or something else entirely, but the new method seems to work.

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -363,11 +363,19 @@ in
               stopScript = pkgs.writeScript "minecraft-stop-${name}" ''
                 #!${pkgs.runtimeShell}
 
-                if ! [ -d "/proc/$1" ]; then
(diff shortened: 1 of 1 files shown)


Infinidoge:
I honestly just took the `$MAINPID` method wholesale from the Terarria server implementation in Nixpkgs. I haven't been actively running Minecraft servers in a bit, so I just never noticed that this was an issue. Thanks for the fix!

Running the simple VM test works locally, so I'll go ahead and try to merge this later today after my exam.

PR #67, opened 2024-06-05
 module: allow management via systemd socket

### LINKED ISSUE #63: Feature Request: Using the module without Tmux
Hi. First things first, thanks for your (and all the contributors') amazing work on this! I spent far too much time packaging various servers & proxies myself, and this project is going to make my life much easier. :heart:

I'd also love to migrate my ad-hoc solution to the provided module, but I don't really like the fact that it runs the server in a tmux session. I love Tmux as a terminal multiplexer, but using it for services has some really annoying problems that I don't think can be solved in any elegant way:
* Logs don't end up in the system journal—this is probably the biggest one. This means that when something goes wrong and the service crashes, I am left without any information whatsoever unless it ends up in the server logs (which, e.g., startup script failures won't.). It also means I can't process the logs automatically via Loki, fail2ban, and other such tools.
* The server can't be nicely automated from the outside. I really don't want to blindly pipe random keys into tmux and pray that they end up in the command prompt, not mixed with some other rubbish.
* Connecting to the socket remotely is unnecessarily complicated.
* Although not strictly forced by using tmux, I dislike having to run all servers under the same user. Ideally, I'd use systemd's `DynamicUser` for every server. Supporting this with Tmux would be fairly complicated.

A solution that I use currently, which has none of these problems, is to just run the server process directly, controlling it via RCON. It's not perfect, particularly because none of the RCON clients I could find are very nice, but I find it to be a much cleaner approach.

I'd love to add support for this workflow to the provided module, but I'm not sure whether such a PR would be accepted. If so, how do you think it should work? A global toggle? A per-server one? Something else? I'm open to all suggestions.

Thanks in advance!

EDIT: I just noticed #52, which addresses most of these issues. I guess I'll use the fork from there for now, and add support for multiple users after it lands.

frantisekhanzlikbl:
Supersedes #52. Partially fixes #63 (per-server user & group configuration missing).

This PR adds a way to make servers managed via a systemd stdin socket and the journal, allowing for a better integrated experience with the rest of the system.

This PR largely builds on top of @Misterio77's excellent #52 and retains the nice additions:
* expanded tests verifying automated command execution
* extra lifecycle hooks for the server services

I did change the default systemd socket path from #52 to be aligned with the current tmux socket path and renamed the lifecycle hooks to preserve the systemd naming convention, though.

I've added a few extras, not all of which I'm sure want to be kept. Feedback welcome. :slightly_smiling_face:
* `docsAsciiDoc` and `docsCommonMark` package outputs (not present in the overlay) to make sure that `nix flake check` verifies that the docs can be built. there was a trivial issue previously, which prevented them from building.
* VS Code config for [nix-ide] specifying the formatter to be `nixpkgs-fmt`, along with the proper indentation.
* per-server management of the socket (both tmux & systemd) paths.

I'm not too sure I like the way server overrides of the global management system config works, but it was all I could conjure with my limited knowledge of advanced usage of the module system.

This PR should be backwards compatible, although some field-testing would be much appreciated as I don't have any real setup to test it on (and frankly no time either :)).

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -176,7 +176,7 @@ let
 
               # Wait for the PID of the minecraft server to disappear before
               # returning, so systemd doesn't attempt to SIGKILL it.
-              tries=3
+              tries=15
               while kill -0 "$1" 2> /dev/null; do
                 if [[ $tries -gt 0 ]]; then
                   sleep 1s
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -152,9 +152,7 @@ let
 
             ${tmux} -S ${sock} send-keys ${escapeShellArg server.stopCommand} Enter
 
-            while server_running ; do
-              sleep 0.25
-            done
+            while server_running; do sleep 1s; done
           '';
         };
       } else if ms.systemd-socket.enable then
@@ -174,18 +172,7 @@ let
             ${optionalString (server.stopCommand != null) ''
               echo ${escapeShellArg server.stopCommand} > ${escapeShellArg (ms.systemd-socket.stdinSocket.path name)}
 
-              # Wait for the PID of the minecraft server to disappear before
-              # returning, so systemd doesn't attempt to SIGKILL it.
-              tries=15
-              while kill -0 "$1" 2> /dev/null; do
-                if [[ $tries -gt 0 ]]; then
-                  sleep 1s
-                else
-                  echo >&2 "Timed out waiting for server to stop."
-                  exit 1
-                fi
-                ((tries--))
-              done
+              while kill -0 "$1" 2> /dev/null; do sleep 1s; done
             ''}
           '';
         };
@@ -679,6 +666,10 @@ in
               serviceConfig = {
                 inherit ExecStartPre ExecStart ExecStartPost ExecStopPost ExecReload;
                 ExecStop = "${execStopScript} $MAINPID";
+
+                # the Minecraft server (as of 1.20.6) has a 60s timeout for saving each world.
+                # let's let it handle potential lock-ups by itself before resorting to killing it.
+                TimeoutStopSec = "1min 15s";
 
                 Restart = conf.restart;
                 WorkingDirectory = "${cfg.dataDir}/${name}";
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -1,4 +1,4 @@
-{ config, lib, pkgs, ... }:
+{ config, lib, options, pkgs, ... }:
 with lib;
 let
   cfg = config.services.minecraft-servers;
@@ -72,6 +72,125 @@ let
     };
   };
 
+  managementSystem = types.submodule {
+    options = {
+      tmux = {
+        enable = mkEnableOption "management via a TMUX socket";
+        socketPath = mkOption {
+          type = with types; functionTo path;
+          description = ''
+            Function from a server name to the path at which the server's tmux socket is placed.
+            To connect to the console, run `tmux -S <path to socket> attach`,
+            press `Ctrl + b` then `d` to detach.
+
+            Note that while currently the default respects <option>services.minecraft-servers.runDir</option>,
+            that option is deprecated and will be removed.
+            The default will then change to `name: "/run/minecraft/''${name}.sock`.
+          '';
+          default = name: "${cfg.runDir}/${name}.sock";
+          defaultText = literalExpression ''name: "''${cfg.runDir}/''${name}.sock"'';
+        };
+      };
+      systemd-socket = {
+        enable = mkEnableOpt "management through the systemd journal & a command socket";
+        stdinSocket = {
+          path = mkOption {
+            type = with types; functionTo path;
+            description = ''
+              Function from a server name to the path at which the server's stdin socket is placed.
+              You can send the server commands by writing to this socket,
+              for example with shell redirection: `echo 'list' > <path to socket>`.
+
+              Note that while currently the default respects <option>services.minecraft-servers.runDir</option>,
+              that option is deprecated and will be removed.
+              The default will then change to `name: "/run/minecraft/''${name}.stdin`.
+            '';
+            default = name: "${cfg.runDir}/${name}.stdin";
+            defaultText = literalExpression ''name: "''${cfg.runDir}/''${name}.stdin"'';
+          };
+          mode = mkOption {
+            type = types.strMatching "[0-7]{4}";
+            description = "Access mode of the socket file in octal notation";
+            default = "0660";
+          };
+        };
+      };
+    };
+  };
+
+  managementSystemConfig = name: server:
+    let
+      ms = server.managementSystem;
+      tmux = "${getBin pkgs.tmux}/bin/tmux";
+    in
+    assert assertMsg (!(ms.tmux.enable && ms.systemd-socket.enable)) "Only one server management system can be enabled at a time.";
+    if ms.tmux.enable then
+      let sock = ms.tmux.socketPath name; in {
+        serviceConfig = {
+          Type = "forking";
+          GuessMainPID = true;
+        };
+        hooks = {
+          start = ''
+            ${tmux} -S ${sock} new -d ${getExe server.package} ${server.jvmOpts}
+
+            # HACK: PrivateUsers makes every user besides root/minecraft `nobody`, so this restores old tmux behavior
+            # See https://github.com/Infinidoge/nix-minecraft/issues/5
+            ${tmux} -S ${sock} server-access -aw nobody
+          '';
+          postStart = ''
+            ${pkgs.coreutils}/bin/chmod 660 ${sock}
+          '';
+          stop = ''
+            function server_running {
+              ${tmux} -S ${sock} has-session
+            }
+
+            if ! server_running ; then
+              exit 0
+            fi
+
+            ${tmux} -S ${sock} send-keys ${escapeShellArg server.stopCommand} Enter
+
+            while server_running ; do
+              sleep 0.25
+            done
+          '';
+        };
+      } else if ms.systemd-socket.enable then
+      {
+        serviceConfig = {
+          Type = "simple";
+          StandardInput = "socket";
+          StandardOutput = "journal";
+          StandardError = "journal";
+        };
+        hooks = {
+          start = ''
+            ${getExe server.package} ${server.jvmOpts}
+          '';
+          postStart = "";
+          stop = ''
+            ${optionalString (server.stopCommand != null) ''
+              echo ${escapeShellArg server.stopCommand} > ${escapeShellArg (ms.systemd-socket.stdinSocket.path name)}
+
+              # Wait for the PID of the minecraft server to disappear before
+              # returning, so systemd doesn't attempt to SIGKILL it.
+              tries=3
+              while kill -0 "$1" 2> /dev/null; do
+                if [[ $tries -gt 0 ]]; then
+                  sleep 1s
+                else
+                  echo >&2 "Timed out waiting for server to stop."
+                  exit 1
+                fi
+                ((tries--))
+              done
+            ''}
+          '';
+        };
+      } else builtins.throw "At least one server management system must be enabled.";
+
   mkEnableOpt = description: mkBoolOpt' false description;
 in
 {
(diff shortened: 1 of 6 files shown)


modules/minecraft-servers.nix
```diff
@@ -101,13 +212,6 @@ in
       the server name in this directory, such as <literal>/srv/minecraft/servername</literal>.
     '';
 
-    runDir = mkOpt' types.path "/run/minecraft" ''
-      Directory to place the runtime tmux sockets into.
-      Each server's console will be a tmux socket file in the form of <literal>servername.sock</literal>.
-      To connect to the console, run `tmux -S /run/minecraft/servername.sock attach`,
-      press `Ctrl + b` then `d` to detach.
-    '';
-
```
Infinidoge, line 110: For the sake of backwards compatibility, this option should be kept, with a warning if it is changed from the default. (And it should accordingly be used in the default tmux socket path function.

modules/minecraft-servers.nix
```diff
@@ -179,8 +297,6 @@ in
 
           restart = mkOpt' types.str "always" ''
             Value of systemd's <literal>Restart=</literal> service configuration option.
-            Due to the servers being started in tmux sockets, values other than
-            <literal>"no"</literal> and <literal>"always"</literal> may not work properly.
```
Infinidoge, line 183: This should be edited to maintain the note that values other than `no` and `always` may not work properly with tmux sockets.

Infinidoge (changes requested):
Overall the implementation looks good! Left some notes with regards to backwards compatibility and documentation. I'll test and do further review on the code later when I have more time.

With that said, something that you should do before this can be merged is split this up into commits with separate concerns. Namely:
- The vscode settings file should be a meta commit
- Documentation export should be its own commit
- Extra pre/post start/stop and custom stop command should be its own commit
- Adding/modifying the tests should be their own commits

frantisekhanzlikbl:
Thanks so much for the review!

I just pushed the requested changes. Please let me know if there is something else to change.

frantisekhanzlikbl:
Hmm, I just tried to migrate my config to this and got bitten in my ass by the extra lifecycle hooks not having `set -euo pipefail`, therefore "succeeding" even if parts of them fail. Would you mind me using `writeShellApplication` instead of `writeShellScript` for them?

This would theoretically be a breaking change, as setups previously "working" might now start failing, but I'd argue that these were broken from the start. I could add an option for that, but I don't think it would be worth the extra code and maintenance.

modules/minecraft-servers.nix
```diff
@@ -102,10 +213,12 @@ in
     '';
 
     runDir = mkOpt' types.path "/run/minecraft" ''
-      Directory to place the runtime tmux sockets into.
+      Deprecated: Directory to place the runtime tmux sockets into.
       Each server's console will be a tmux socket file in the form of <literal>servername.sock</literal>.
       To connect to the console, run `tmux -S /run/minecraft/servername.sock attach`,
       press `Ctrl + b` then `d` to detach.
+
+      Plase use <option>services.minecraft-servers.managementSystem.tmux.socketPath</option>` instead.
```
Infinidoge, line 221: This should have a corresponding [module system warning](https://nixos.org/manual/nixos/unstable/#sec-assertions). This would work:
```
warnings = lib.optional (cfg.runDir != "/run/minecraft") ''
  `runDir` has been deprecated.

  Please use `services.minecraft-servers.managementSystem.tmux.socketPath` instead.
  For example, `name: "${cfg.runDir}/''${name}.sock"`

  See the changelog file for more information.
'';
```
Otherwise end users who don't check this repo every so often won't get the message.

modules/minecraft-servers.nix
```diff
@@ -72,6 +72,117 @@ let
...
+  managementSystem = types.submodule {
+    options = {
+      tmux = {
+        enable = mkEnableOption "management via a TMUX socket";
+        socketPath = mkOption {
+          type = with types; functionTo path;
+          description = ''
+            Function from a server name to the path at which the server's tmux socket is placed.
+            To connect to the console, run `tmux -S <path to socket> attach`,
+            press `Ctrl + b` then `d` to detach.
```
Infinidoge, line 84: Could you add a note here that the default will change to be `name: "/run/minecraft/${name}.sock`" after `runDir` gets removed?

modules/minecraft-servers.nix
```diff
@@ -72,6 +72,117 @@ let
...
+      };
+      systemd-socket = {
+        enable = mkEnableOpt "management through the systemd journal & a command socket";
+        stdinSocket = {
+          path = mkOption {
+            type = with types; functionTo path;
+            description = ''
+              Function from a server name to the path at which the server's stdin socket is placed.
+              You can send the server commands by writing to this socket,
+              for example with shell redirection: `echo 'list' > <path to socket>`.
```
Infinidoge, line 98: Same as previous comment, would like a note stating that the default will be changing, just so its documented.

modules/minecraft-servers.nix
```diff
@@ -72,6 +72,117 @@ let
...
+      };
+    };
+  };
+
+  managementSystemConfig = name: server:
+    let
+      ms = server.managementSystem;
+      tmux = "${getBin pkgs.tmux}/bin/tmux";
+    in
+    assert assertMsg (!(ms.tmux.enable && ms.systemd-socket.enable)) "Only one server management system can be enabled at a time.";
```
Infinidoge, line 118: This should be done with a [module system assertion](https://nixos.org/manual/nixos/unstable/#sec-assertions) instead. Which actually, I'm pretty sure can be handled in the `managementSystem` submodule type, which would be clean.

modules/minecraft-servers.nix
```diff
@@ -72,6 +72,117 @@ let
...
+                else
+                  echo >&2 "Timed out waiting for server to stop."
+                  exit 1
+                fi
+                ((tries--))
+              done
+            ''}
+          '';
+        };
+      } else builtins.throw "At least one server management system must be enabled.";
```
Infinidoge, line 184: Same as previous, this would be best handled with a module system assertion.

Infinidoge (changes requested):
The single period change in CHANGELOG looks out of place on the diff, but also I don't think it's big enough to warrant its own commit so it can stay.

Something that I'm a bit mixed about is how which system to use is selected. Needing to explicitly turn off the default and turn on the one you want feels... wrong to me, so an enum option like
```nix
{
  system = "tmux"
  tmux.socketPath = name: ...;
}
```
might be better. This would also alleviate the need for assertions to prevent selecting 2 at the same time, with the cost of it being not idiomatic. If anyone has any thoughts about this, please chime in and tell me my code conciseness perfectionism is on the loose again.

With regards to pipefail, yeah go right ahead. These kinds of scripts should fail early, and is the kind of breaking change that should lead to more fixed problems than actual problems.

Some other small comments about module assertions and docs. Haven't tested it myself yet, but will review again after I do.

modules/minecraft-servers.nix
frantisekhanzlikbl: ouch, I totally forgot about that. sorry, I'll add it shortly. :sweat_smile:

modules/minecraft-servers.nix
frantisekhanzlikbl: I thought about that, but for some reason decided against it. I honestly don't remember why, and if it works at the submodule level, I don't see why not.

frantisekhanzlikbl: Oh, I know. Doing the validation as a module system assertion at the submodule level prevents the user from having `{ tmux.enable = false; systemd-socket.enable = false; }` in the global config and then choosing for each server. It might be too niche of a use case, but I can very much see myself doing this.

modules/minecraft-servers.nix
frantisekhanzlikbl: Even if this one is handled via the module system assertions, this throw must stay here too, as we really don't have any meaningful return value if no management system is enabled and returning an arbitrary one feels really wrong.

frantisekhanzlikbl:
Alright, I just pushed the next iteration. Sorry about the dot change in the changelog, that might've been a little too pedantic of me. I updated the descriptions of socket path options and moved to `writeShellApplication`.

I thought about using a string enum for the preferred management system too, but decided against it, as it would make it non-practical to add more management systems (such as RCON), that don't conflict with others. i.e. both systemd-socket and tmux could be used along with RCON, just not with each other.

I kept the language-level assertions for now, for the reasons described in my comment on the appropriate review comments, but I don't feel that strongly about the first one. If you do, I can change that.

modules/minecraft-servers.nix
Infinidoge: The module system assertion would force at least 1 management system to be enabled, since it would refuse to finish evaluating otherwise.

frantisekhanzlikbl: I believe that is only true for evaluating entire NixOS configurations. i.e. things that evaluate the top-level system attribute[^1]. Any uses of the module system that do not need to do that would not inform the user of anything, and just silently produce nonsense. a good example would be manually evaluating a part of a NixOS config in the REPL during debugging, which I do fairly often.

[^1]: https://github.com/NixOS/nixpkgs/blob/db4bd243ca966a89a2bead2f9275d9163268d856/nixos/modules/system/activation/top-level.nix#L66

Infinidoge: Fair. We'll just leave it as is then/

modules/minecraft-servers.nix
Infinidoge: Makes sense. It feels unclean, but doesn't really need to be perfect, and it works as is.

modules/minecraft-servers.nix
```diff
@@ -323,6 +499,15 @@ in
         }
       ];
 
+      warnings = lib.optional (cfg.runDir != options.services.minecraft-servers.runDir.default) ''
```
Infinidoge, line 502: Hmm. How does this interact with `mkDefault`? If someone used `mkDefault`, then this might not warn them of the change.

frantisekhanzlikbl: I believe that `mkDefault` is just a normal value just like any other for the module system and does not affect the default at all. There doesn't seem to be anything special about its implementation: https://github.com/NixOS/nixpkgs/blob/bc3cfcb6ed2167858598d316683f1d40ca322898/lib/modules.nix#L1014-L1020

Infinidoge: Ah yep, just an override with priority, nevermind!

frantisekhanzlikbl:
I just remembered that systemd has its own options for timeouts, so I pushed a new commit that lets systemd handle server stop timeouts instead of the ad-hoc implementation that was in place previously in the systemd-socket management system (and completely missing in the tmux one).

marek-rychly:
What is the status of this merge request? Can it be merged?

Logging into journald is much more better than tmux for monitoring a server, so I am using the last commit from this merge request now. Unfortunately, there are newer fabric versions in the master branch which are not available here. Rebase or merge to the master would be appreciated. Thank you.

frantisekhanzlikbl:
> What is the status of this merge request? Can it be merged?
> 
> Logging into journald is much more better than tmux for monitoring a server, so I am using the last commit from this merge request now. Unfortunately, there are newer fabric versions in the master branch which are not available here. Rebase or merge to the master would be appreciated. Thank you.

I believe all the brought up concerns have been resolved on my side.

I've been using the latest commit here on a few small-scale servers in production for about three months and have faced no problems, but my deployments are also probably fairly simple.

Currently, I'm severely short on time, so if @Infinidoge has more change requests before this is mergeable, the PR will most likely have to wait for a while or get taken up by somebody else. I'll do a rebase in the meantime, though.

modules/minecraft-servers.nix
```diff
@@ -72,6 +72,112 @@ let
...
+              ${tmux} -S ${sock} has-session
+            }
+
+            if ! server_running ; then
+              exit 0
+            fi
+
+            ${tmux} -S ${sock} send-keys ${escapeShellArg server.stopCommand} Enter
+
+            while server_running; do sleep 1s done
```
Infinidoge, line 155:
```suggestion
            while server_running; do sleep 1s; done
```

Infinidoge (approved):

Infinidoge:
Note to self: GitHub action that runs `nix flake check`

modules/minecraft-servers.nix
frantisekhanzlikbl: ouch, dumb mistake. sorry and thanks for catching it. :)

frantisekhanzlikbl:
thanks for the reviews and merge, it is greatly appreciated! :)

modules/minecraft-servers.nix
Infinidoge: No problem :)

I should definitely add CI since this could have been caught sooner by just having GH run `nix flake check`. It's on my I'll Do It Eventually list

Misterio77:
Thanks so much for picking this up! Looks awesome.

PR #80, opened 2024-07-15
build-support: Update asm to version 9.5

NikSneMC:
Compatibility level "JAVA_21" can't be applied with asm older than 9.5


NikSneMC:
> I question why this was done as 3 separate commits, but they'll be squashed when I merge so that isn't a problem.
> 
> Has this been tested to also work for older Textile versions using Java 17? (Fabric 1.14, for example)

This was done as 3 commits because it was experiment "What minimal version will work?". I can squash them. 

I haven't tested it on older version yet, but I don't think something could break there

mechan1sm:
Shouldn't the ASM version be set dynamically instead of being hardcoded? (#83)

PR #81, opened 2024-07-18
module: add operators

CartConnoisseur:
Adds a new option, `servers.<name>.operators`, to generate `ops.json`, similarly to `whitelist.json`.

It redefines `minecraftUUID` which doesn't feel very good, but to be completely honest I'm pretty new to nix and don't know where a wider-scoped definition would fit better

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -226,6 +226,52 @@ in
             '';
           };
 
+          operators = mkOption {
+            type =
+              let
+                minecraftUUID = types.strMatching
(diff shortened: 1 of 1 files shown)


Infinidoge:
For future reference, please rebase when syncing with upstream, instead of merging in changes from upstream. Using a rebase instead of a merge keeps the commit history clean. This is less of a problem, because I squash-merge PRs, but it is preferable nonetheless. I am starting my break between semesters, so I'll try and review this properly soon.

Infinidoge (approved):

PR #84, opened 2024-07-24
Dynamically set the asm version

mechan1sm:
Added the small workaround to extract the asm version from list of libaries to the [mkTextileLoader.nix](https://github.com/mechan1sm/nix-minecraft/blob/master/pkgs/build-support/mkTextileLoader.nix#L19).

The list of libraries must contain exactly one string matching the following regular expression:
```regex
org\.ow2\.asm:asm:([\.0-9]+)
```
This should fix those errors for fabric/legacy-fabric/quilt servers, cause they are only ones using mkTextileLoader.

### DIFF
--- a/pkgs/build-support/mkTextileLoader.nix
+++ b/pkgs/build-support/mkTextileLoader.nix
@@ -16,6 +16,9 @@
 let
   lib_lock = lib.importJSON ./libraries.json;
   fetchedLibraries = lib.forEach libraries (l: fetchurl lib_lock.${l});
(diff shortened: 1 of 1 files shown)


mechan1sm:
Also, during testing I found that [quilt loader 0.17.5-beta.4 has broken versions in its loader_locks.json](https://github.com/Infinidoge/nix-minecraft/blob/master/pkgs/quilt-servers/loader_locks.json#L2866).

Issue #83, opened 2024-07-24
Hardcoded Implementation-Version in manifest


mechan1sm:
As described in [fabric-servers/loader_locks](https://github.com/Infinidoge/nix-minecraft/blob/master/pkgs/fabric-servers/loader_locks.json#L953), fabric-loader 0.16.0 uses ASM 9.6.

There is a code in [mkTextileLoader.nix](https://github.com/Infinidoge/nix-minecraft/blob/d70e2db81e5a36a085609cf735e74cdac161cb5b/pkgs/build-support/mkTextileLoader.nix#L36), that has hardcoded
```
Implementation-Version: 9.2
```

This causes some mods such as carpet-tis-addition for mc1.21 crash the server. (probably related to #79)

I already tested this with hardcoded
```
Implementation-Version: 9.6
```
and it didn't crash.

I'm not sure if hardcoding the `Implementation-Version` is a good idea.
Looks like it should be added into specific lock files.

Issue #79, opened 2024-06-30
Errors when starting server


greysilly7:
Please help me I cant figure it out

code
> {
  pkgs,
  lib,
  ...
}: let
  modpack = pkgs.fetchPackwizModpack {
    url = "https://github.com/Alpaca-Industries/iremia-comeback/raw/1.0.0/pack.toml";
    packHash = "sha256-i+W0WA/eH/tNmYzguP2/v/rJb1LUnCV4WmMO9FpjcdI=";
  };
  mcVersion = modpack.manifest.versions.minecraft;
  fabricVersion = modpack.manifest.versions.fabric;
  serverVersion = lib.replaceStrings ["."] ["_"] "fabric-${mcVersion}";
in {
  services.minecraft-servers.servers.iremia = {
    enable = true;
    autoStart = true;
    package = pkgs.fabricServers.${serverVersion}.override {loaderVersion = fabricVersion;};

    serverProperties = {
      server-port = 18611;
      /*
      difficulty = 3;
      gamemode = 1;
      max-players = 5;
      motd = "IREMIA!";
      white-list = false;
      */
      enable-rcon = true;
      "rcon.password" = "password";
    };

    symlinks = {
      "mods" = "${modpack}/mods";
      # "config" = "${modpack}/config";
    };
  };
}

errors:
> Jun 30 15:28:01 greyserver wqnygk7snd6hzyswf2aigwv0pi6yazpn-minecraft-start-iremia[192689]: error connecting to /run/minecraft/iremia.sock (Permission denied)
Jun 30 15:28:01 greyserver wqnygk7snd6hzyswf2aigwv0pi6yazpn-minecraft-start-iremia[192690]: error connecting to /run/minecraft/iremia.sock (Permission denied)
Jun 30 15:28:01 greyserver systemd[1]: minecraft-server-iremia.service: Control process exited, code=exited, status=1/FAILURE`

> × minecraft-server-iremia.service - Minecraft Server iremia
     Loaded: loaded (/etc/systemd/system/minecraft-server-iremia.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Sun 2024-06-30 15:28:02 EDT; 5min ago
    Process: 192667 ExecStartPre=/nix/store/zmf8qdrgvavixb95gfdf27957cai4560-unit-script-minecraft-server-iremia-pre-start/bin/minecraft-server-iremia-pre-start (code=exited, status=0/SUCC>
[...]

PR #85, opened 2024-07-24
textile-servers: dynamically set jar asm version

mechan1sm:
Added the small workaround to extract the asm version from list of libaries to the [mkTextileLoader.nix](https://github.com/mechan1sm/nix-minecraft/blob/master/pkgs/build-support/mkTextileLoader.nix#L19).

The list of libraries must contain exactly one string matching the following regular expression:
```nix
"org\\.ow2\\.asm:asm:([\.0-9]+)"
```
This should fix those errors for fabric/legacy-fabric/quilt servers, cause they are only ones using mkTextileLoader.

Related to issues #79 and #83, and pull request #80

### DIFF
--- a/pkgs/build-support/mkTextileLoader.nix
+++ b/pkgs/build-support/mkTextileLoader.nix
@@ -14,11 +14,11 @@
 }:
 
 let
+  inherit (builtins) head filter map match;
+  
   lib_lock = lib.importJSON ./libraries.json;
   fetchedLibraries = lib.forEach libraries (l: fetchurl lib_lock.${l});
-  asmVersion = builtins.head (builtins.head (
-    builtins.filter (v: v!=null) (builtins.map (builtins.match "org\\.ow2\\.asm:asm:([\.0-9]+)") libraries)
-  ));
+  asmVersion = head (head (filter (v: v!=null) (map (match "org\\.ow2\\.asm:asm:([\.0-9]+)") libraries)));
 in
 stdenvNoCC.mkDerivation {
   pname = "${loaderName}-server-launch.jar";
--- a/pkgs/build-support/mkTextileLoader.nix
+++ b/pkgs/build-support/mkTextileLoader.nix
@@ -15,10 +15,10 @@
 
 let
   inherit (builtins) head filter map match;
-  
+
   lib_lock = lib.importJSON ./libraries.json;
   fetchedLibraries = lib.forEach libraries (l: fetchurl lib_lock.${l});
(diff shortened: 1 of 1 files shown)


mechan1sm:
Also, during testing I found that [quilt loader 0.17.5-beta.4 has broken versions in its loader_locks.json](https://github.com/Infinidoge/nix-minecraft/blob/master/pkgs/quilt-servers/loader_locks.json#L2866).
(Tested only version evaluation for fabric/fabric-legacy/quilt servers. The only one that errors is quilt 0.17.5-beta.4)

pkgs/build-support/mkTextileLoader.nix
```diff
@@ -16,6 +16,9 @@
 let
   lib_lock = lib.importJSON ./libraries.json;
   fetchedLibraries = lib.forEach libraries (l: fetchurl lib_lock.${l});
+  asmVersion = builtins.head (builtins.head (
+    builtins.filter (v: v!=null) (builtins.map (builtins.match "org\\.ow2\\.asm:asm:([\.0-9]+)") libraries)
+  ));
```
Infinidoge, line 21:
```suggestion
let
  inherit (builtins) head filter map match;
  
  lib_lock = lib.importJSON ./libraries.json;
  fetchedLibraries = lib.forEach libraries (l: fetchurl lib_lock.${l});
  asmVersion = head (head (filter (v: v!=null) (map (match "org\\.ow2\\.asm:asm:([\.0-9]+)") libraries)));
```

Bit of cleanup, all of the `builtins` was just noise.

Infinidoge (changes requested):
Haven't tested the packages of this myself yet, just a small code style nitpick. Overall looks good to me. I'm not the biggest fan of using regex to solve this, and think this is the kind of thing that should be handled in the update scripts/lock files, however this works as a quick fix that just solves the problem, especially without needing to process the lockfiles to backfill information.

Infinidoge:
Also as a note with regards to #84, I squash pull requests, so you didn't need to recreate the PR to turn it into 1 commit. (Could also rebase/squash yourself, but that takes git knowledge most people avoid lol)

Infinidoge:
Please avoid merging the master branch into PR branches, and rebase instead. It clutters the history and can get messy with squash merging.

mechan1sm:
Ah, github "sync fork" feature performs merge instead of rebase. Will you merge this PR? Should I revert this merge commit and rebase, or do something else?

Infinidoge:
Go ahead and rebase on master, and it should theoretically remove the merge commits automatically.

Infinidoge (approved):

Infinidoge:
Running the VM tests before merging.

PR #98, opened 2024-08-30
meta: document overriding custom server jar

JimmJam:
Some server jar files are either too niche to be merged into the project, or make my head spin on even attempting to get officially merged with the Python scripting involved to do so. However, I've found it's very easy to override the src of the vanilla server's package to support custom jar files. My small contribution is documenting how to do so and letting people know it's possible :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,8 @@ All of these packages are also available under `packages`, not just `legacyPacka
 - `velocity-server`: Same as `velocityServers.velocity`
 - `minecraft-server`: Same as `vanilla-server`
 
-Server versions not found above can be manually added via an override.
+Server versions not found above can be manually via an override.
+This override changes the path the launch script to your provided jar file, and does not modify the vanilla jar.
 In this example, uberbukkit.jar exists next to this configuration:
 
 ```nix
--- a/README.md
+++ b/README.md
@@ -208,9 +208,8 @@ All of these packages are also available under `packages`, not just `legacyPacka
 - `velocity-server`: Same as `velocityServers.velocity`
 - `minecraft-server`: Same as `vanilla-server`
 
-Server versions not found above can be manually via an override.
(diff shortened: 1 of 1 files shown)


Infinidoge:
Good catch!

That said, to me this feels like a small hole in the libraries that could be improved on. Something like a `buildServer` helper that accepts a `jar` argument, perhaps.

Infinidoge (changes requested):
Add a note clarifying that this sets the path to the jar in a launch script (and not actually modifying a vanilla server), and this should be good to go

JimmJam (commented):
Added the comment about overriding path, not jar

README.md
```diff
@@ -208,6 +208,24 @@ All of these packages are also available under `packages`, not just `legacyPacka
 - `velocity-server`: Same as `velocityServers.velocity`
 - `minecraft-server`: Same as `vanilla-server`
 
+Server versions not found above can be manually via an override.
```
Infinidoge, line 211:
```suggestion
Server versions not found above can be setup manually via an override.
```

README.md
```diff
@@ -208,6 +208,24 @@ All of these packages are also available under `packages`, not just `legacyPacka
...
+This override changes the path the launch script uses to your provided jar file, and does not modify the vanilla jar.
+In this example, uberbukkit.jar exists next to this configuration:
```
Infinidoge, line 213:
```suggestion
For example, this override changes the path the launch script uses to your provided jar file, and does not modify the vanilla jar:
```

I don't think it's necessary to specify that the jar is alongside, since that should be implicit in how the code is written.

Infinidoge (changes requested):
Couple small nitpicks (which you can commit right from GitHub, and this is good to merge :)

At some point I'll probably add a build tool for this so it doesn't need to be done with an override, but that's a later problem.

PR #103, opened 2024-10-08
txt-file generation & docs symlinks/files value examples

arminius-smh:
Hello,

two commits, one for the option to automatically infer and generate txt files, another for expanding the docs a bit on how
`symlinks.value` and `files.value` work.

I am not sure if `Every option and example in symlinks works the same way for files` is actually correct? It looked like it from the definitions of the options anyhow.
Just, when I was reading the docs, I didn't understand that I could just exchange it 1:1 and get the files instead of symlinks, since it just said they are "similiar", I didn't quite know what to do with that, which is I tried to clarify that a bit and added another example to the files.


Misterio77 (approved):
LGTM

Misterio77:
Can we merge this one, @Infinidoge?

Infinidoge:
For future reference, it's best if PRs are 1 thing only, as it makes the final commit history cleaner :)

arminius-smh:
Sorry, yes ofc, I just did the changes concurrently for my issue and didn't think of it as two changes, which I also closed now, since this solution is good enough for me now

PR #116, opened 2024-12-26
refactor file & symlink

### LINKED ISSUE #70: Document that the module's `files` option does env substitution
Currently, neither the readme nor the option docs mention that `services.minecraft-servers.servers.<name>.files` performs environment substitution.

I just got bitten by this when working with binaries that were getting mysteriously corrupted, and the module mutating them was not something that I expected.

The ideal case scenario is IMHO that this is disabled by default and opt-in on each file, or at least opt-able out of, but it should really be documented. I can make a PR for either if it helps. :slightly_smiling_face:

### LINKED ISSUE #73: Allow the files option to accept directories
Forgive my ignorance if this isn't possible, but it would be nice if the files option could accept directories.

The symlinks option does not work for the config folder since it needs to be writable (this is mentioned in the files option docs). I'd rather not specify every config file individually in the files option 😂

Misterio77:
Hi there! I was still bothered by the fact that symlinks/files aren't cleaned up when removed from the config (and stopWhenChanged prevents us from using stuff such as reload), so I decided to try a different approach.

The new system works by adding a file (.nix-minecraft-managed) to track managed files and symlinks. This makes it possible to correctly cleanup managed files (when stopping, restarting, reloading).

I've also added support for directories (https://github.com/Infinidoge/nix-minecraft/issues/73), and ensured only non-binary files are substituted with env vars (https://github.com/Infinidoge/nix-minecraft/issues/70)

Marked as draft as I'm still testing it.

Fixes #73
Fixes #70

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -602,45 +602,62 @@ in
 
             msConfig = managementSystemConfig name conf;
 
+            markManaged = file: ''echo "${file}" >> .nix-minecraft-managed'';
+            cleanAllManaged = ''
+              if [ -e .nix-minecraft-managed ]; then
+                readarray -t to_delete < .nix-minecraft-managed
+                rm -rf "''${to_delete[@]}"
+                rm .nix-minecraft-managed
+              fi
+            '';
+
+
             ExecStartPre =
               let
+                backup = file: ''
+                  if [[ -e "${file}" ]]; then
+                    echo "${file} already exists, moving"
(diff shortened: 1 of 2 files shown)


Misterio77:
Fixes #73
Fixes #70

Misterio77:
It seems that extended attributes don't work with symlinks :(

I switched my approach to keeping the managed files listed in a text file

Infinidoge (approved):

Issue #92, opened 2024-08-08
Generate `allowed_symlinks.txt` when symlinks are used in module


Infinidoge:
See the [1.20 Pre-release 7 patch notes](https://feedback.minecraft.net/hc/en-us/articles/16224502352141-Minecraft-Java-Edition-1-20-Pre-release-7) and [corresponding help article](https://help.minecraft.net/hc/en-us/articles/16165590199181).

The naïve solution would be to do
```
symlinks = {
  "allowed_symlinks.txt" = pkgs.writeText "allowed_symlinks.txt" "/nix/store";
};
```
An ideal solution would be
```
alowedSymlinks = [ "/nix/store" ];
```

This would solve the problem brought up in #74.

Misterio77:
This should work too (format from 7408a562f548ba6a523058f819f3a8e429b1783a), and has all the module-system niceties:

```
symlinks = {
  "allowed_symlinks.txt" = {
    value = ["/nix/store"];
    format = {
      type = lib.types.listOf str;
      generate = name: value: pkgs.writeText name (lib.concatStringsSep "\n" value);
    };
  };
};
```

But a convenience `allowedSymlinks` would be nice.

PR #117, opened 2024-12-26
add allowedSymlinks option

Misterio77:
Fixes https://github.com/Infinidoge/nix-minecraft/issues/92.

Draft as I still got to test it.


Misterio77:
Test case added, and ready to go!

modules/minecraft-servers.nix
```diff
@@ -449,6 +457,24 @@ in
             '';
           };
 
+          allowedSymlinks = mkOption {
+            default = ["/nix/store"];
```
Infinidoge, line 461: Rather than setting this as a default, it should instead be added as part of `config`. Otherwise, it leads to the unintuitive behavior of adding to `allowedSymlinks` removing `/nix/store` from the list. Removing `/nix/store` should require `lib.mkForce`.

Infinidoge (changes requested):
Minor nitpick.

Fix that and rebase on master to remove the cherry pick and this should be good to merge.

modules/minecraft-servers.nix
Misterio77: You're totally right. I'll get that fixed

nyabinary: bump :p

Infinidoge:
Please also rebase on master to remove the cherrypick

modules/minecraft-servers.nix
Misterio77: Sorry for the delay. Just starting my yearly minecraft phase, so time to fix my nix-minecraft stuff!

modules/minecraft-servers.nix
```diff
@@ -567,6 +586,10 @@ in
                 example = options.services.minecraft-servers.managementSystem.example;
               };
             };
+
+            config = mkIf config.enable {
+              allowedSymlinks = [ "/nix/store" ];
+            };
```
Infinidoge, line 592:
```suggestion

            config = {
              allowedSymlinks = [ "/nix/store" ];
            };
```
(And remove the `config` reference at the top of the section)

I don't think this being set will have any side effects if the server isn't enabled to begin with, so having it here makes it kinda redundant. Plus, it would prevent seeing the value of `allowedSymlinks` if the server is disabled. (Though that doesn't really matter in practice.)

If there is a notable reason why this needs to be gated under `enable` though, then it can stay.

Infinidoge (changes requested):
Looks great! One small nit, plus the commit should have the right name format (though I can fix this myself when I rebase-merge)

PR #129, opened 2025-02-14
Purpur init

### LINKED ISSUE #37: Package Purpur servers
Package [Purpur](https://purpurmc.org/), based on the [Paper server packaging](https://github.com/Infinidoge/nix-minecraft/tree/master/pkgs/paper-servers), with additional reference to the [nix-minecraft-servers implementation](https://github.com/jyooru/nix-minecraft-servers/blob/main/packages/purpur/default.nix).

heisfer:
resolves  #37

## Things Done
- [x] Updated all purpur builds lock file
- [x] Updated Readme   
- [x]  `nix fmt`
- [x]  `nix flake check` 
- [x] Tested `pkgs.purpurServers.purpur` in own config
## Test code
```nix
{ inputs, pkgs, ... }:
{
  imports = [ inputs.nix-minecraft.nixosModules.minecraft-servers ];
  nixpkgs.overlays = [ inputs.nix-minecraft.overlay ];

    enable = true;
    eula = true;
    servers = {
      purpur = {
        enable = true;
        package = pkgs.purpurServers.purpur;
      };
    };
  };
}

```

## Notes
- Purpur api doesn't store sha256
- Package is modified version of paper package
- `nix fmt` formatted other unrelated files

### DIFF
--- a/pkgs/purpur-servers/default.nix
+++ b/pkgs/purpur-servers/default.nix
@@ -1,0 +1,63 @@
+{ callPackage
+, lib
+, jdk8
+, jdk11
+, jdk
+, vanillaServers
+,
+}:
+let
+  inherit (lib.our) escapeVersion;
+  inherit (lib)
+    nameValuePair
+    flatten
+    last
+    versionOlder
+    mapAttrsToList
+    ;
+  versions = lib.importJSON ./lock.json;
+
+  # Remove -build... suffix
+  stripBuild = v: builtins.head (builtins.match "(.*)-build.*" v);
+  # Sort by attribute 'attr' using 'f' function
+  sortBy = attr: f: builtins.sort (a: b: f a.${attr} b.${attr});
+
+  # https://docs.papermc.io/paper/getting-started#requirements
+  getRecommendedJavaVersion =
+    v:
+    if versionOlder v "1.11.2" then
+      jdk8
+    else if versionOlder v "1.16.5" then
+      jdk11
+    else
+      jdk;
+
+  packages = mapAttrsToList
+    (
+      mcVersion: builds:
+        sortBy "version" versionOlder (
+          mapAttrsToList
+            (
+              buildNumber: value:
+                callPackage ./derivation.nix {
+                  inherit (value) url sha256;
+                  version = "${mcVersion}-build.${buildNumber}";
+                  jre = getRecommendedJavaVersion mcVersion;
+                  minecraft-server = vanillaServers."vanilla-${escapeVersion mcVersion}";
+                }
+            )
+            builds
+        )
+    )
+    versions;
+
+  # Latest build for each MC version
+  latestBuilds = sortBy "version" versionOlder (map last packages);
+in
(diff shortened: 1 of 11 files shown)


Infinidoge:
Please revert all formatting changes, they cause a lot of noise in an otherwise clean and straight forward PR.

Additionally, is there any reason why the update script is written to be async? I can't really see any reason why it should be, and it adds unnecessary complexity.

heisfer:
> Please revert all formatting changes, they cause a lot of noise in an otherwise clean and straight forward PR.

Sure thing!

> Additionally, is there any reason why the update script is written to be async? I can't really see any reason why it should be, and it adds unnecessary complexity.

The reason for that is fastly download old build and generate sha256 hash (since they only support md5 and I believe md5 is deprecated in nix). I'll revert it back to non async since it's no longer needed.

heisfer:
Things changed:
1. Reverted `update.py`
2. Added bad.json so we don't request not found builds each time.
3. Rechecked java versions.
4. Added `log4j.xml` for specific versions. 
5. Formated with `nixfmt-rfc-style` as it says in `flake.nix` (CONTRIBUTING.MD says nixpkgs-fmt)

I hope commit message is fit for [CONTRIBUTING.md](https://github.com/Infinidoge/nix-minecraft/blob/master/CONTRIBUTING.md)

pkgs/purpur-servers/default.nix
```diff
@@ -0,0 +1,71 @@
...
+  sortBy = attr: f: builtins.sort (a: b: f a.${attr} b.${attr});
+
+  # https://docs.papermc.io/paper/getting-started#requirements
+  getRecommendedJavaVersion =
+    v:
+    # oldest version is 1.14.1
+    # Version older than 1.1 1.1 = false
+    if versionOlder v "1.17" then
+      jdk11
+    else if versionOlder v "1.18.2" then # paper says 1.18.1+ but 1.18.1 max is 17
```
Infinidoge, line 32: The behavior doesn't seem to match the comment

Infinidoge (commented):
Overall looks good to me, small question with regards to the java version selector

pkgs/purpur-servers/default.nix
heisfer: jar files seems to have hard limit on max version
`Unsupported Java detected (61.0). Only up to Java 16 is supported.` - 1.16.5
`Unsupported Java detected (65.0). Only up to Java 17 is supported.` - 1.18.1 
So I'll leave it to that.

heisfer: Another note is that 1.17 version seems to be completly broken. 
Doesn't support jdk11 and 17 is unsupported.

fmbearmf:
any news on the status of this?

Infinidoge:
Apologies for taking so long to get back to this. Would it be possible to skip including the `url` in the lockfile, since the URL is predictable based on the Minecraft version and the build number? That would cut down on a lot of lines in the lockfile.

Issue #65, opened 2024-05-29
Classpath broken on Fabric


otomir23:
Looks like https://github.com/Infinidoge/nix-minecraft/commit/fa037f484d8775119a5e4fb3f20aaec15c19c118 makes Fabric Loader have two instances on the classpath at once, which is an [illegal state](https://github.com/FabricMC/fabric-loader/blob/1a833267b54beea5eb635222df4af149f8a6a1d6/src/main/java/net/fabricmc/loader/impl/util/LoaderUtil.java#L50). The server fails to start. 

<details><summary>Stack trace for reference (fabric-1.20.1-0.15.11)</summary>

```
java.lang.IllegalStateException: trying to load net.fabricmc.loader.impl.FabricLoaderImpl from target class loader
        at net.fabricmc.loader.impl.util.LoaderUtil.verifyNotInTargetCl(LoaderUtil.java:55) ~[16h8ywfxn9dhw6mci6cz40yglyvp456r-fabric-loader-0.15.11.jar:?]
        at net.fabricmc.loader.impl.FabricLoaderImpl.<clinit>(FabricLoaderImpl.java:619) ~[16h8ywfxn9dhw6mci6cz40yglyvp456r-fabric-loader-0.15.11.jar:?]
        at net.fabricmc.loader.api.FabricLoader.getInstance(FabricLoader.java:42) ~[16h8ywfxn9dhw6mci6cz40yglyvp456r-fabric-loader-0.15.11.jar:?]
        at net.minecraft.class_6396.mdf82ea6$fabric-crash-report-info-v1$lambda$fillSystemDetails$0$1(class_6396.java:544) ~[server-intermediary.jar:?]
        at net.minecraft.class_6396.method_37123(class_6396.java:66) ~[server-intermediary.jar:?]
        at net.minecraft.class_6396.handler$zzm000$fabric-crash-report-info-v1$fillSystemDetails(class_6396.java:541) ~[server-intermediary.jar:?]
        at net.minecraft.class_6396.<init>(class_6396.java:58) ~[server-intermediary.jar:?]
        at net.minecraft.class_128.<init>(class_128.java:34) ~[server-intermediary.jar:?]
        at net.minecraft.class_128.method_24305(class_128.java:274) ~[server-intermediary.jar:?]
        at net.minecraft.server.Main.main(Main.java:102) ~[server-intermediary.jar:?]
[...]

PR #130, opened 2025-02-18
Fix textile servers

### LINKED ISSUE #60: Quilt issue: quilt_loader missing
There seem to be an issue with the Quilt loader. Here is my config: https://github.com/sweenu/nixfiles/blob/d6fba43e8177dd67d9ba7ddb769fa934d8ebb4a2/hosts/najdorf/minecraft/default.nix
And here is the crash report: [crash-2024-03-20_24.14.02.7060-quilt_loader.txt](https://github.com/Infinidoge/nix-minecraft/files/14657851/crash-2024-03-20_24.14.02.7060-quilt_loader.txt)

I thought at first that it might be the modpack even though I have no problem using it client side. But then I tried removing all mods and keeping only the Quilt API mod and the error persisted. So I'm thinking that it might be an issue with the package.

jackwilsdon:
This PR reverts 1ba686c30955166aad2d46077d86aa81e9010eec and adds some extra changes on top to fix Fabric and some runtime dependency detection issues. Fixes #60, #65.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -92,5 +92,25 @@ lib.makeExtensible (
       mapListToAttrs (x: builtins.unsafeDiscardStringContext (lib.removePrefix "${path}/" x)) (lib.id) (
         lib.filesystem.listFilesRecursive "${path}"
       );
+
+    wrapJarManifest =
+      manifestText:
+      let
+        chunkCharacters' =
+          characters: chunks:
+          if
+            characters != [ ] # 71 characters plus space prefix = 72 line length
+          then
+            chunkCharacters' (drop 71 characters) (chunks ++ [ (take 71 characters) ])
+          else
+            chunks;
+
+        chunkCharacters =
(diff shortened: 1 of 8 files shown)


jackwilsdon:
Tested that the following start properly:
* `quiltServers.quilt-1_19_4`
* `quiltServers.quilt-1_20_6`
* `quiltServers.quilt-1_21_4`
* `fabricServers.fabric-1_19_4`
* `fabricServers.fabric-1_20_6`
* `fabricServers.fabric-1_21_4`


You can verify that Nix knows about the package's runtime dependencies using the following commands:
```shell
$ nix build .#quiltServers.quilt-1_19_4
$ nix-store --query --requisites result | grep jar$
```

Infinidoge:
Went ahead and squashed the commits together into one with a couple changes, including doing the libraries fix on all new libraries.

Thank you very much for the fixes! ^^

Issue #142, opened 2025-05-19
on-demand start/stop server when players join/leave (lazymc?)


Yeshey:
possibly add an option to add `lazymc.enable = true` and `lazymc.config = {...}`

[lazymc](https://github.com/timvisee/lazymc) already exists in nixpkgs.

if I end up doing this for myself I might make a PR

PR #145, opened 2025-05-20
Add lazymc (on-demand) functionality to nix-minecraft

Yeshey:
Solves https://github.com/Infinidoge/nix-minecraft/issues/142
This adds the options:
```nix
        lazymc = {
          enable = true;
          package = pkgs.lazymc;
          config = { ... };
        };
```
Under each instance. This makes servers go to sleep when there are no players and wake up when one player tries to join [lazymc](https://github.com/timvisee/lazymc)

#### `servers.<name>.lazymc`

Integrates [lazymc](https://github.com/timvisee/lazymc), putting server to sleep when idle and waking it upon player connection.

*   **`enable`**: `boolean`, default `false`

*   **`package`**: `package`, default `pkgs.lazymc`  
    The `lazymc` package to use. You might have to change lazymc version according to your minecraft server version, for example lazymc v0.2.10 supports Minecraft Java Edition 1.7.2+, while for Minecraft Java Edition 1.20.3+ you'll need lazymc v0.2.11

*   **`config`**: `attribute set`, default `{}`  
    Allows defining and overriding settings in the `lazymc.toml`, for all options see [here](https://github.com/timvisee/lazymc/blob/master/res/lazymc.toml). The auto generated config options are:
    *   `server.command`: Automatically set to use the server `package` and `jvmOpts`;
    *   `server.directory`;
    *   `server.address`: Automatically set to `127.0.0.1:<serverProperties.server-port or 25565>";`

    **Firewall**
    *   When `lazymc.enable = true`, the  `openFirewall` option for this server instance will open the port specified in `lazymc.config.public.address` (or its default), not the internal Minecraft `serverProperties.server-port`.

    **Example:**
    ```nix
    { pkgs, ... }: {
        eula = true;
        servers.myLazyServer = {
          enable = true;
          package = pkgs.paperServers.paper-1_18_2;
          serverProperties = {
            "server-port" = 25566;
            "max-tick-time" = -1; # Recommended with lazymc
          };
    
          lazymc = {
            enable = true;
            package = let
            # you can use https://lazamar.co.uk/nix-versions/
              pkgs-with-lazymc_0_2_10 = import (builtins.fetchTarball {
                  url = "https://github.com/NixOS/nixpkgs/archive/336eda0d07dc5e2be1f923990ad9fdb6bc8e28e3.tar.gz";
                  sha256 = "sha256:0v8vnmgw7cifsp5irib1wkc0bpxzqcarlv8mdybk6dck5m7p10lr";
              }) { inherit (pkgs) system; };
            in pkgs-with-lazymc_0_2_10.lazymc;
            config = {
              public.address = "0.0.0.0:25565";
            };
          };
        };
      };
    }
    ```

### DIFF
--- a/modules/minecraft-servers.nix
+++ b/modules/minecraft-servers.nix
@@ -151,6 +151,11 @@ let
     let
       ms = server.managementSystem;
       tmux = "${getBin pkgs.tmux}/bin/tmux";
+      runCommand =
+        if server.lazymc.enable then
+          "${server.lazymc.package}/bin/lazymc start --config lazymc.toml"
+        else
+          "${getExe server.package} ${server.jvmOpts}";
     in
     assert assertMsg (
       !(ms.tmux.enable && ms.systemd-socket.enable)
@@ -166,7 +171,7 @@ let
         };
         hooks = {
           start = ''
-            ${tmux} -S ${sock} new -d ${getExe server.package} ${server.jvmOpts}
+            ${tmux} -S ${sock} new -d ${runCommand}
 
             # HACK: PrivateUsers makes every user besides root/minecraft `nobody`, so this restores old tmux behavior
             # See https://github.com/Infinidoge/nix-minecraft/issues/5
@@ -175,7 +180,7 @@ let
           postStart = ''
             ${pkgs.coreutils}/bin/chmod 660 ${sock}
           '';
-          stop = ''
+          stop = optionalString (!server.lazymc.enable) ''
             function server_running {
               ${tmux} -S ${sock} has-session
             }
@@ -199,12 +204,10 @@ let
           StandardError = "journal";
         };
         hooks = {
-          start = ''
-            ${getExe server.package} ${server.jvmOpts}
-          '';
+          start = runCommand;
           postStart = "";
           stop = ''
-            ${optionalString (server.stopCommand != null) ''
+            ${optionalString (server.stopCommand != null && !server.lazymc.enable) ''
               echo ${escapeShellArg server.stopCommand} > ${escapeShellArg (ms.systemd-socket.stdinSocket.path name)}
 
               while kill -0 "$1" 2> /dev/null; do sleep 1s; done
@@ -634,6 +637,52 @@ in
                 default = { };
                 example = options.services.minecraft-servers.managementSystem.example;
               };
+              lazymc = {
+                enable = mkEnableOpt "Puts your Minecraft server to rest when idle, and wakes it up when players connect.";
+                package = mkOption {
+                  type = types.package;
+                  default = pkgs.lazymc;
+                  defaultText = literalExpression "pkgs.lazymc";
+                  description = "The lazymc package to use.";
+                };
+                config = mkOption {
+                  type = types.submodule {
+                    freeformType = types.attrs;
+                    options = {
+                      public = mkOption {
+                        type = types.submodule {
+                          freeformType = types.attrs;
+                          options = {
+                            address = mkOption {
+                              type = types.strMatching "^[^:]+:[0-9]+$";
+                              default = "0.0.0.0:25565";
+                              description = "Public address in format 'address:port'";
+                            };
+                          };
+                        };
+                        default = { };
+                      };
+                    };
+                  };
+                  default = { };
+                  description = ''
+                    Configuration for lazymc, mirroring its TOML structure.
+                    Some values like server.command, server.directory and server.address will
+                    be automatically configured by this module when generating lazymc.toml. 
+
+                    See <link xlink:href="https://github.com/timvisee/lazymc/blob/master/res/lazymc.toml"/>
+                    for documentation on these values.
+                  '';
+                  example = literalExpression ''
+                    {
+                      public.address = "0.0.0.0:25565"; # Public port for players
+                      time.sleep_after = 900; # Sleep after 15 minutes
+                      motd.sleeping = "Shhh, the server is napping!";
+                      server.forge = false; # Set to true if this is a Forge server
+                    }
+                  '';
+                };
+              };
             };
 
             config = {
@@ -648,6 +697,17 @@ in
   config = mkIf cfg.enable (
     let
       servers = filterAttrs (_: cfg: cfg.enable) cfg.servers;
+
+      getPublicPort =
+        conf:
+        if conf.lazymc.enable then
+          let
+            addr = conf.lazymc.config.public.address or "0.0.0.0:25565";
+            portStr = lib.last (lib.splitString ":" addr);
+          in
+          lib.toInt portStr
+        else
+          conf.serverProperties.server-port or 25565;
     in
     {
       users = {
@@ -680,15 +740,35 @@ in
         {
           assertion =
             let
-              serverPorts = mapAttrsToList (name: conf: conf.serverProperties.server-port or 25565) (
+              serverPorts = mapAttrsToList (name: conf: getPublicPort conf) (
                 filterAttrs (_: cfg: cfg.openFirewall) servers
               );
 
               counts = map (port: count (x: x == port) serverPorts) (unique serverPorts);
             in
             lib.all (x: x == 1) counts;
-          message = "Multiple servers are set to use the same port. Change one to use a different port.";
+          message = "Multiple servers are set to use the same public port (either through lazymc public.address or server.properties server-port). Change one to use a different port.";
         }
+        (
+          let
+            lazymcPortConflicts = filter (x: x.hasConflict) (
+              mapAttrsToList (name: conf: {
+                inherit name;
+                hasConflict =
+                  conf.lazymc.enable && getPublicPort conf == (conf.serverProperties.server-port or 25565);
+              }) servers
+            );
+            conflictingServerNames = map (x: x.name) lazymcPortConflicts;
+          in
+          {
+            assertion = lazymcPortConflicts == [ ];
+            message = ''
+              Server(s): ${concatStringsSep ", " conflictingServerNames} has the same port set for serverProperties.server-port and lazymc.config.public.address
+              Please set for example: `serverName.serverProperties.server-port = 25566;`, lazymc's internal server.address will automatically point to it
+              Lazymc's public.address is "0.0.0.0:25565" by default
+            '';
+          }
+        )
       ];
 
       warnings = lib.optional (cfg.runDir != options.services.minecraft-servers.runDir.default) ''
@@ -706,7 +786,9 @@ in
           # Minecraft and RCON
           getTCPPorts =
             n: c:
-            [ c.serverProperties.server-port or 25565 ]
+            [
+              (getPublicPort c)
+            ]
             ++ (optional (c.serverProperties.enable-rcon or false) (c.serverProperties."rcon.port" or 25575));
           # Query
           getUDPPorts =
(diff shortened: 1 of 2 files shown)


Yeshey:
Maybe it would have been better to not use the server-port - 1 and just let the user do all the config themeselves, there is quite a bit of logic bc of that, but there would always be some abstraction because openFirewall would have to automatically open the port in lazymc instead of in serverproperties if it was active. And I do think that having a one click solution like this is more user friendly as long as we can make it work well

Infinidoge:
Please avoid closing and opening pull requests. You can force push to your branch to edit commits.

Infinidoge:
With regards to the actual content of the PR:
- TOML writing can be handled through the `files` option of the server, as opposed to reimplementing it yourself. See [here](https://github.com/Infinidoge/nix-minecraft/blob/850d167ecbde1f1a7ac86b491a3494d439d4976b/modules/minecraft-servers.nix#L690-L705) for an example
- The lazymc config overwrites the management system config, which it **should never do**. Based on how you wrote it, it seems that lazymc should instead be a management system in and of itself, as right now it incorrectly overwrites load-bearing parts of the existing management systems, breaking them.
- Overall the implementation feels... very verbose and overcomplicated.

Yeshey:
Thanks for reviewing, I didn't realize that deleting and renaming my repo would also close my PRs, I'll try to keep further changes in this one.
Yeah, the code isn't ideal, when I did it i was probably thinking about my own needs more than doing something clean to try to merge, and might have used an LLM to help a bit 😅.
Sorry if it wasted your time, soon as I get a bit of time I'll try to rewrite it. On a rewrite I think I'd drop the `"server-port" - 1` logic, `openFirewall` would open the `lazymc` port instead of the mc server one if lazymc is enabled; and just let the user manage the ports themeselves, idk if you want to weigh in on that

Titaniumtown:
Dont use a llm for anything. opens the project up to legal issues.

CalmSeoul:
Will this eventually be merged with the main branch? This would be an awesome addition cause I'm having a tough time trying to get lazymc working with nix-minecraft manually, but having it built in would be incredible

Yeshey:
@CalmSeoul this is a clunky implementation, as it stands it should be rewriten to get merged. I will try to look at it in the summer but i cant promise. I'm not the most proficiant nixer as well.
If u want to use it rn, you can import the flake from my fork and check the readme for how its set up.

Yeshey:
Rewrote the implementation, sorry for the fuss earlier, no more vibe code now, hopefully it can get merged

Titaniumtown:
I tested this and it works! Good job.

Only comment is that, shouldn't lazymc's public port be 25565 and the internal port be 25566? It should be transparent to the end user and not be another port imo.

Yeshey:
I thought picking another port would be kinda arbitrary, so I decided to leave the assertion to force the user to always pick their lazymc public port instead. Maybe you're right that that should suggest port 25565, and now that I'm thinking about that, there should probably be another assertion for when the internal server port and the lazymc public port are the same

Yeshey:
There you go, I think this is better

now if a user goes

```nix
        servers.testerido = {
          enable = true;
          lazymc.enable = true;
        };
```

they'll get:
```
       error: Server 'testerido' has the same port set for serverProperties.server-port and lazymc.config.public.address
       Please set for example: `testerido.serverProperties.server-port = 25566;`, lazymc's internal server.address will automatically point to it
       Lazymc's public.address is "0.0.0.0:25565" by default
```

Titaniumtown (changes requested):
Just a few comments, mostly good!

modules/minecraft-servers.nix
```diff
@@ -178,7 +183,18 @@ let
               exit 0
             fi
 
-            ${tmux} -S ${sock} send-keys ${escapeShellArg server.stopCommand} Enter
+            ${
+              if server.lazymc.enable then
+                ''
+                  # Won't gracefully stop if server running unfortunately
```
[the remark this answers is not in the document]
Titaniumtown: > I was trying to send a INT signal directly to lazymc but it looks like that's [not fully implemented on their side yet](https://github.com/timvisee/lazymc/blob/master/src/service/signal.rs?rgh-link-date=2025-08-18T17%3A15%3A47Z#L30)

Hm ok. I think this is not expected behavior from an end user then. This is very scary!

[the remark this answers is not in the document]
Yeshey: I just confirmed this also works with `managementSystem.systemd-socket.enable = true;`, no exception at all, and stops gracefully. The above exception must be because tmux died witch takes the server down with it witch is what we want (https://github.com/PaperMC/Velocity/issues/1439#issuecomment-2381329081)

I'm going to remove that comment then as it doesn't make sense anymore

Being stopped like this is basically how lazymc's expecting to be stopped anyways so I think it's working as expected, and any other errors would probably be down to lazymc, so I think this is resolved then

modules/minecraft-servers.nix
```diff
@@ -690,6 +759,13 @@ in
                 bypassesPlayerLimit = v.bypassesPlayerLimit;
               }) conf.operators;
               "server.properties".value = conf.serverProperties;
+              "lazymc.toml".value = lib.recursiveUpdate {
+                server = {
+                  command = "${getExe conf.package} ${conf.jvmOpts}";
+                  directory = ".";
+                  address = "127.0.0.1:${toString conf.serverProperties.server-port or 25565}";
+                };
+              } conf.lazymc.config;
```
Infinidoge: This will create the `lazymc.toml` file even if not configured. Please add a `mkIf` guard.

The other values don't have a `mkIf` guard as their values will be an empty attribute set (or otherwise empty value) when not configured.

Infinidoge (changes requested):
If I am honest, I don't really want this functionality to be baked into the module. This is something the user can theoretically manage for themself. However considering the complexity of that, I am not (yet) rejecting this outright.

With that said, in addition to the comments on files, ***please*** rebase out all automated lockfile commits, and change your branch to something other than master. Automated commits pollute the PR. (Note to self, add that to CONTRIBUTING.md)

Yeshey:
Cool thanks! I rebased, but instead of renaming the branch (that would close the PR and we would loose the discussion) I disabled github actions on my fork, hopefully that prevents the clutter, I'll go through the file comments soon as I can

modules/minecraft-servers.nix
Yeshey: Couldn't chuck a mkIf in there, ended up solving it like this:

```nix
            // (optionalAttrs conf.lazymc.enable {
              "lazymc.toml".value = lib.recursiveUpdate {
                server = {
                  command = "${getExe conf.package} ${conf.jvmOpts}";
                  directory = ".";
                  address = "127.0.0.1:${toString conf.serverProperties.server-port or 25565}";
                };
              } conf.lazymc.config;
            })
```

modules/minecraft-servers.nix
```diff
@@ -184,7 +189,18 @@ let
               exit 0
             fi
 
-            ${tmux} -S ${sock} send-keys C-u ${escapeShellArg server.stopCommand} Enter
+            ${
+              if server.lazymc.enable then
+                ''
+                  # Won't gracefully stop if server is running unfortunately
```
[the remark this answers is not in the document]
Infinidoge: Mmm I wouldn't rely on `ps` and `pgrep`, especially based on name. `ps` based on PID would be fine.

rharish101:
Is there anything else left to do? I'd like to use this without manually using @Yeshey's fork.

Yeshey:
> Is there anything else left to do? I'd like to use this without manually using @Yeshey's fork.

I'll try to explain the situation as it is right now
The way this nixOS module stops the Minecraft servers is by directly sending the letters "stop" (or any other keyword to stop the server according to the type of server) into the stdin of tmux so the server stops itself, and when the server stops,  _only then_ tmux will also exit and the systemd service will exit as well. 
But this is not how lazymc works, lazymc sends a `SIGINT` (or `SIGTERM`?) signal to the server and hopes that the server stops correctly independently of the type of server it is. And it doesn't expose any API so we can't easily know if the server has already shut down or not, so what enabling lazymc now does is it kills tmux (witch might bring up non fatal exception on the minecraft server because the terminal suddenly dissapears) that in turn kills lazymc witch takes care of killing the server as well with signals.
This to say that lazymc handles server shutdowns fundamentally differently than this module, and accepting that would mean that all the logic for stopping servers by sending stop words is kinda irrelevant at least when lazymc is enabled. 
We could theoretically check if the server process is suspended or running to see if it is sleeping or not initialized and work around this, but this is complex and I think this is a pretty bad idea as lazymc is already expecting to handle server shutdown with signals.

I see why there's friction, but if @Infinidoge sees a path for this to get merged I'd also love to be able to finally get this done

modules/minecraft-servers.nix
```diff
@@ -566,6 +611,17 @@ in
...
       servers = filterAttrs (_: cfg: cfg.enable) cfg.servers;
+
+      getPublicPort =
+        _: conf:
+        if conf.lazymc.enable then
+          let
+            addr = conf.lazymc.config.public.address or "0.0.0.0:25565";
+            portStr = lib.last (lib.splitString ":" addr);
+          in
+          lib.toInt portStr
```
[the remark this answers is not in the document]
Infinidoge: Not overkill. This can be checked in the NixOS option's type through type guards. I believe `lib.types.strMatching` (or whatever it's actually called) is what you want.

modules/minecraft-servers.nix
```diff
@@ -200,11 +207,11 @@ let
         };
         hooks = {
           start = ''
-            ${getExe server.package} ${server.jvmOpts}
+            ${runCommand}
           '';
```
Infinidoge: Replace with
```
start = runCommand
```

modules/minecraft-servers.nix
```diff
@@ -566,6 +602,17 @@ in
   config = mkIf cfg.enable (
     let
       servers = filterAttrs (_: cfg: cfg.enable) cfg.servers;
+
+      getPublicPort =
+        _: conf:
```
Infinidoge:
```suggestion
      getPublicPort =
        conf:
```

modules/minecraft-servers.nix
```diff
@@ -598,15 +645,35 @@ in
         {
           assertion =
             let
-              serverPorts = mapAttrsToList (name: conf: conf.serverProperties.server-port or 25565) (
+              serverPorts = mapAttrsToList (name: conf: getPublicPort _ conf) (
```
Infinidoge:
```suggestion
              serverPorts = mapAttrsToList (name: conf: getPublicPort conf) (
```

modules/minecraft-servers.nix
```diff
@@ -624,7 +691,9 @@ in
           # Minecraft and RCON
           getTCPPorts =
             n: c:
-            [ c.serverProperties.server-port or 25565 ]
+            [
+              (getPublicPort n c)
```
Infinidoge:
```suggestion
              (getPublicPort c)
```

Infinidoge (changes requested):

Infinidoge:
Sorry for the lack of responses, currently midterm season at university so I haven't been able to be active on GitHub.
As such it will likely be a while longer before this will be merged.

samiser:
@Infinidoge just to make sure i understand, by a "separate module" do you mean a distinct NixOS module still within nix-minecraft (like how https://github.com/Infinidoge/nix-minecraft/pull/150 is adding a separate home-manager module)? or something else?

samiser:
i've been working on [a fork](https://github.com/samiser/nix-minecraft/tree/lazymc-module) that implements lazymc as a separate module, but in testing i've found an issue where lazymc fails to detect the server starting (see https://github.com/timvisee/lazymc/issues/89). i've found that a workaround is to use an older version (0.2.10) but i'm confused how @Yeshey your fork works, did you test it using papermc? maybe that's my issue

in any case this is a bit awkward since i've had to provide an overlay for the older package. unsure if there's some other workaround for this that would avoid having to pin and overlay this version, i spent some time digging into it but didn't find anything.

regardless of that, is my implementation more in line with what you were thinking @Infinidoge? i added an example of usage in examples/ in my fork, but you can also see my real usage of it [in my personal configs](https://github.com/samiser/nix-configs/blob/main/hosts/minecraft/minecraft-server.nix)

PR #146, opened 2025-05-27
Make fetchPackwizModpack work with an `src` argument

### LINKED ISSUE #110: fetchPackwizModpack support parameter src as drv instead of url
I hope these code

```nix
  modpack = pkgs.fetchPackwizModpack {
    url = "https://github.com/Misterio77/Modpack/raw/0.2.9/pack.toml";
    packHash = "sha256-L5RiSktqtSQBDecVfGj1iDaXV+E90zrNEcf4jtsg+wk=";
  };
```
  
  Can be replaced into:

```nix
  modpack = pkgs.fetchPackwizModpack {
    src = fetchFromGithub { owner = "Misterio77"; repo = "Modpack"; rev = "master"; sha256 = "";};
  };
```

And if the name of `pack.toml` is not special enough, we can specify it like

```nix
  modpack = pkgs.fetchPackwizModpack {
    src = fetchFromGithub { owner = "Misterio77"; repo = "Modpack"; rev = "master"; sha256 = "";};
    path = "pack.toml";
  };
```

ehllie:
Allows for setting an src argument for the `fetchPackwizModpack` function. It needs to be a derivation or a file path containing a `pack.toml` file.

I'm using `updog` to serve the contents of `src` to `packwizInstallerBootstrap` rather than `packwiz serve`, since the latter would attempt to update the contents of the working directory on top of just serving the files, which does not work for read only nix store file paths.

I've been using the code [for myself](https://github.com/ehllie/vanilla-plus/blob/main/bootstrapPackwiz.nix) for a while now, and figured I'd try to upstream that change.

Should close #110.

### DIFF
--- a/pkgs/tools/fetchPackwizModpack/default.nix
+++ b/pkgs/tools/fetchPackwizModpack/default.nix
@@ -12,13 +12,13 @@
 let
   fetchPackwizModpack =
     {
-      pname ? "packwiz-pack",
-      version ? "",
-      url,
+      # Provide a path to the packwiz modpack source directory,
+      # or a URL to a pack.toml file.
+      src ? null,
+      url ? null,
       packHash ? "",
       # Either 'server' or 'both' (to get client mods as well)
       side ? "server",
-
       # The derivation passes through a 'manifest' expression, that includes
       # useful metadata (such as MC version).
       # By default, if you access it, IFD will be used. If you want to use
@@ -27,6 +27,19 @@ let
       manifestHash ? null,
       ...
     }@args:
+    let
+      srcNull = src == null;
+      urlNull = url == null;
+      toml = builtins.fromTOML (builtins.readFile (src + "/pack.toml"));
+      pname = args.pname or (if !srcNull then toml.name else "packwiz-pack");
+      version = args.version or (if !srcNull then toml.version else "");
+      drv = fetchPackwizModpack args;
+      bootstrapUrl = if !urlNull then url else "file://${src}/pack.toml";
+    in
+
+    assert lib.assertMsg (
+      srcNull != urlNull # equivalent of (src != null) xor (url != null)
+    ) "Either 'src' or 'url' must be provided to fetchPackwizModpack";
 
     stdenvNoCC.mkDerivation (
       finalAttrs:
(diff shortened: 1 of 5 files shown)


pkgs/tools/fetchPackwizModpack/default.nix
```diff
@@ -57,15 +74,39 @@ let
...
-          java -jar "$packwizInstallerBootstrap" \
-            --bootstrap-main-jar "$packwizInstaller" \
-            --bootstrap-no-update \
-            --no-gui \
-            --side "${side}" \
-            "${url}"
-        '';
+        buildPhase = lib.concatLines [
+          ''
+            set -euo pipefail
```
ehllie, line 79: I've configured to use stricter bash evaluation settings, since for a reason I could not figure out, the build would forever hang if `src` was ever malformed, for example by only having `pack.toml` but no `mods` directory referenced inside the `pack.toml` file. The [packwizInstaller](https://github.com/Infinidoge/nix-minecraft/pull/146/files#diff-52f36499ca7677199e307fb2b571d509b015529cfbcfd5ad2cd118eff0730238R100-R105) would then fail, but the build would just wait forever, and the builder would continue to hold onto the derivation lock even after a console interrupt.

ehllie:
> This looks good to me. It would be nice to add a test case, though.

I've added a test file, though admittedly it's my first time doing that. I also had a hard time making sure it worked properly since I don't have access to a linux machine to see if a failed build results in a failed test.

Infinidoge:
packwiz-installer supports `file://` URLs, meaning that you can skip using `updog` entirely.

ehllie:
> packwiz-installer supports `file://` URLs, meaning that you can skip using `updog` entirely.

Thanks! That is indeed a much more elegant solution than running a dummy web server!

tests/packwiz-from-src/default.nix
```diff
@@ -0,0 +1,23 @@
...
+  name = "packwiz-from-src-check";
+  doCheck = true;
+  phases = [
+    "checkPhase"
+    "installPhase"
+  ];
+  checkPhase = ''
+    set -euo pipefail
+    test -f ${pack}/pack.toml
+  '';
```
Infinidoge, line 21: Could you add a bit more to the check phase to make sure the packwiz pack downloaded the mod correctly?

Infinidoge (changes requested):
Everything pretty much looks good!

Besides adding a bit more to the test, please change the commit messages to match the [commit message format](https://github.com/Infinidoge/nix-minecraft/blob/master/CONTRIBUTING.md#commit-message-format) (notably not conventional commits)

tests/packwiz-from-src/default.nix
ehllie: I have a slight issue with running the tests. It's somewhat unclear to me as to how I should run them on a `aarch64-darwin` system I'm using currently, and even when I enable this test to run on systems other than linux inside the flake file, `nix test` passes even should I add an `exit 1` line inside the check phase.

ehllie: I've added a check for a presence of a `.jar` file inside the `mods` folder. Not sure if that's enough of a check, but I also am not sure how to check that more thoroughly.

Infinidoge (approved):

Issue #136, opened 2025-03-30
[Feature Request]Provide home manager module


snylonue:
The current nix-minecraft implementation uses nixos modules, requiring root privileges for system-level configuration, which is not possible in some cases.

With home-manager module, we can deploy minecraft servers without root privileges.

PR #150, opened 2025-06-03
feat: add home-manager module

nathanregner:
Related: #136

# Summary of commits

## feat: add home-manager module

All options are supported minus `openFirewall`, `user`, `group`. The `dataDir` and `runDir` options are not shared, as they have slightly different defaults

**Notes:**

- Perhaps we could share even more code here if we didn't use NixOS's abstracted service/socket options. `home-manager` doesn't help you out much here.

- Systemd hardening options seem to break the tmux tests specifically. I didn't spend much time digging into this, so they're only enabled for the NixOS module.

## refactor: factor out common config/refactor: add nixos- prefix to file names

For ease of reviewing changes, this was done before adding the home-manager module. Just moving code around for reusability.

To "prove" nothing changed, I recommend diffing these two commits with `--color-moved`.

I also checked the output of vm tests with `nix-diff`:

```sh
nom build .\#checks.x86_64-linux.nixos-simple.driver.nodes.server.system.build.toplevel
nix diff result ../master/result
...
```

## bump: flake inputs

Update to NixOS 25.05 so we don't have to pin an old version of `home-manager`


modules/home/minecraft-servers.nix
```diff
@@ -0,0 +1,158 @@
...
+    pipe
+    types
+    ;
+
+in
+{
+  imports = [ ../shared/minecraft-servers.nix ];
+
+  options.services.minecraft-servers = {
+    dataDir = mkOpt' types.str "%h/srv/minecraft" ''
```
Infinidoge, line 29: I'm not sure about this as the default location. A more appropriate place would probably be `XDG_DATA_HOME` (`%D`) or `XDG_STATE_HOME` (`%S`), as `srv` isn't a 'standard' location in the home directory. (Plus using `minecraft-servers` instead of `minecraft` to avoid conflicts.)

`.local/share/minecraft-servers/whatever` is a bit lengthy, but I think makes more sense as a default.

Infinidoge (changes requested):
Few miscellaneous things:
- I'd prefer it if the library stuff was moved to the root `/lib`, however if it is too annoying to implement then that can be a refactor I'll look into later.
- Rename `shared` to `common`. I just think it sounds better.
- Note that commit messages are **not** conventional commits for this repo. See [CONTRIBUTING](https://github.com/Infinidoge/nix-minecraft/blob/master/CONTRIBUTING.md) for more information.

flake.lock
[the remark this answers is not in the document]
nathanregner: Now that I'm looking: you're right the `flake-compat` bump was totally unnecessary.

However (I wasn't very clear about this), the `nixpkgs` bump was to avoid the following warning:

```
nom build .\#checks.x86_64-linux.home-manager-simple.driver

evaluation warning: minecraft profile: You are using

                      Home Manager version 25.11 and
                      Nixpkgs version 25.05.

                    Using mismatched versions is likely to cause errors and unexpected
                    behavior. It is therefore highly recommended to use a release of Home
                    Manager that corresponds with your chosen release of Nixpkgs.

                    If you insist then you can disable this warning by adding

                      home.enableNixpkgsReleaseCheck = false;

                    to your configuration.
```

Though I could pin an old version of `home-manager` (or just tell it to ignore the warnings in the tests)

modules/home/minecraft-servers.nix
nathanregner: Agreed

nathanregner:
>     * I'd prefer it if the library stuff was moved to the root `/lib`, however if it is too annoying to implement then that can be a refactor I'll look into later.

I can totally do that. I avoided doing so to prevent `mkOpt'`, `mkBoolOpt'`, etc from becoming part of the public API. Honestly it was probably a bit overkill pulling those functions into their own files.

>     * Rename `shared` to `common`. I just think it sounds better.

Agreed

>     * Note that commit messages are **not** conventional commits for this repo. See [CONTRIBUTING](https://github.com/Infinidoge/nix-minecraft/blob/master/CONTRIBUTING.md) for more information.

Fixed (I think, wasn't exactly sure what do with the module refactors)

flake.lock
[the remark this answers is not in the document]
Infinidoge: I'll go ahead and bump Nixpkgs on master so you can rebase on top and bump just home manager. Probably should do that anyways.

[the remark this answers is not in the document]
Infinidoge: Bumped in https://github.com/Infinidoge/nix-minecraft/commit/862e62f46d27681b2195904e18e3b670efa81b24, feel free to rebase

.gitignore
```diff
@@ -1,4 +1,6 @@
-**/.mypy_cache
 **/__pycache__
```
Infinidoge, line 1: I think mypy cache was accidentally removed

modules/common/minecraft-servers.nix
```diff
@@ -0,0 +1,567 @@
...
+    mkService
+    ;
+
+  mkReadOnlyOption =
+    default:
+    mkOption {
+      internal = true;
+      readOnly = true;
+      inherit default;
+    };
```
Infinidoge, line 41: This can be moved into `lib`

modules/common/minecraft-servers.nix
```diff
@@ -0,0 +1,567 @@
...
+            assert assertMsg (
+              !(ms.tmux.enable && ms.systemd-socket.enable)
+            ) "Only one server management system can be enabled at a time.";
+            if ms.tmux.enable then
+              ms.tmux._config
+            else if ms.systemd-socket.enable then
+              ms.systemd-socket._config
+            else
+              builtins.throw "At least one server management system must be enabled.";
+        };
```
Infinidoge, line 204: Is it possible to change this into a module system assertion? Using asserts like this halts early, instead of being collected together with other assertions.

modules/common/minecraft-servers.nix
```diff
@@ -0,0 +1,567 @@
...
+              };
+
+              whitelist = mkOption {
+                type =
+                  let
+                    minecraftUUID =
+                      types.strMatching "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
+                      // {
+                        description = "Minecraft UUID";
+                      };
```
Infinidoge, line 332: While this file is being reorganized, could you move this up to the top-level? Right now it's redefined multiple times, when it shouldn't be.

modules/home/minecraft-servers.nix
```diff
@@ -0,0 +1,156 @@
...
+              Description = "Minecraft Server ${name}";
+              Requires = socket;
+              PartOf = socket;
+              After = [ "network.target" ] ++ socket;
+              StartLimitIntervalSec = 120;
+              StartLimitBurst = 5;
+            };
+
+            Service = service.serviceConfig // {
+              # re-implement NixOS systemd.services.<name>.<environment/path>
```
Infinidoge, line 132: Weird that this isn't already in home-manager tbh. One would think there would be some demand for this.

Infinidoge (changes requested):
Overall looks great! Couple small things, plus I have a few notes-to-self for potential refactors down the line, but nothing that stands in the way of this.

modules/common/minecraft-servers.nix
nathanregner: Good call; done

Infinidoge:
Sorry for the lack of activity, uni and life have taken me off of GitHub for a while.

Code looks good, commits are clean, very good PR. Only thing missing is a bit of documentation. Could you document the home module a bit in the readme? (I really need to setup a docs website...)

Make sure there is a note about `loginctl enable-linger`. Without linger, the Minecraft server will be shut off after logging out of ssh, etc, which would be unexpected behavior.

nathanregner:
> Sorry for the lack of activity, uni and life have taken me off of GitHub for a while.
> 
> Code looks good, commits are clean, very good PR. Only thing missing is a bit of documentation. Could you document the home module a bit in the readme? (I really need to setup a docs website...)
> 
> Make sure there is a note about `loginctl enable-linger`. Without linger, the Minecraft server will be shut off after logging out of ssh, etc, which would be unexpected behavior.

No worries whatsoever! Added docs but they may be a bit sparse; I can add more if you think it'd be helpful

samiser:
any news on potentially merging this? would be great if this could be merged so we could look at working further on #145

PR #159, opened 2025-06-27
flake: addition of overlay.nix in tree

Rexcrazy804:
this lets non flake users manually import the flake's overlay (skipping flake-compat)
```nix
nixpkgs.overlays = [(import "${nix-minecraft}/overlay.nix")];
```


Infinidoge:
While I do think this is a reasonable thing to have, I'd rather not implement it in this way, as it duplicates the code between the flake and the overlay.nix

PR #162, opened 2025-07-17
fix hashes in examples

gdw2:
the sha512's were wrong (truncated?) and nix still complained with the full version.  256 works fine.


Infinidoge:
Good catch, they were truncated! I went ahead and got the full sha512 hashes from `nix-modrinth-prefetch`, as that felt more "correct" for the example (as `nix-modrinth-prefetch` emits sha512 hashes directly from Modrinth's API, as opposed to generating them in Nix.

PR #168, opened 2025-09-01
tools/minecraftctl: init

### LINKED ISSUE #166: CLI for interacting with the server
the simplest implementation will be a bash 

## Proposed API

```sh
# not decided on how to implement it, using env vars to pass the information? or list /srv/minecraft?
minecraftctl list # list available instances with their status

# socket interaction
minecraftctl send <instance> <command> # send command to the tmux session
minecraftctl tail <instance> [-f] # tail logs, with optional follow flag

# systemd management
minecraftctl status <instance> # show status of the instance
minecraftctl stop <instance> # pause the instance
minecraftctl start <instance> # start the paused instance
minecraftctl restart <instance> # restart it
```

aster-void:
This PR adds a CLI Tool for managing and monitoring nix-minecraft servers.

closes #166

## Features

| feature                                       | description                              |
| --------------------------------------------- | -----------------------------------      |
| list                                          | list available nix-minecraft instances   |
| status [instance]                              | show status of the instance              |
| tail [instance] [-fF] [--retry] [-n length] | tail the log of the instance             |
| send [instance] [command]                     | send command to the instance             |
| reload [instance]                             | send reload command to the instance      |
| start [instance]                              | start the instance                       |
| stop  [instance]                              | stop the instance                        |
| restart [instance]                            | restart the instance                     |
| uuid [player]                                 | (utility) fetches UUID of the player     |

## System Expectation

it expects:
- env NIX_MINECRAFT_MINECRAFTCTL_FILE to be set to the path to the file below
- file minecraftctl.json that describes nix-minecraft configuration. see spec.md for expected formatting.


Infinidoge (changes requested):
I'm not personally sure if a `minecraftctl` is within scope for nix-minecraft, so I'm not sure if this will be accepted as a feature in general. If there are examples of management interfaces like this in Nixpkgs, please let me know where.

With regards to the implementation itself:
- Avoid relying on and parsing names as "canonical" information. Instead, add needed metadata to the packages themselves. `getPackageInfo` should be replaced with relevant metadata in `passthru` on the original derivations.
- I *really* do not think bash is the right language for this. I will grant, this is in part due to my dislike of anything moderately complex being written in bash. Needing to reinvent several wheels is... not great.
- Instead of using an environment variable, create a symlink in `/run/minecraft/servers.json` as a static path. Since this requires a rebuild anyways, 
- I don't think this needs a spec.

aster-void:
@Infinidoge 
Thank you for the review!

I found this as an example of adding a CLI helper for easier maintenance:

https://search.nixos.org/options?channel=unstable&show=services.borgbackup.jobs
> Adding a job will cause a borg-job-NAME wrapper to be added to your system path, so that you can perform maintenance easily. See also the chapter about BorgBackup in the NixOS manual.

as well as agenix (although agenix is not in nixpkgs).

For the implementation part, I'm rewriting it in python.

PR #171, opened 2025-09-06
Clear input in tmux session before sending stop command

fliplus:
This pull request clears any input in the tmux session before sending the stop command, preventing stop failures if leftover text is present.


Infinidoge:
Good catch!

For future reference, please use the commit message format in the [CONTRIBUTING file](https://github.com/Infinidoge/nix-minecraft/blob/master/CONTRIBUTING.md).

PR #173, opened 2025-09-07
Add operators option documentation to README.md

fliplus:
This pull request adds documentation for the `servers.<name>.operators` option in the `README.md` file.


Infinidoge:
Surprised this went undocumented so long, good catch!

PR #195, opened 2026-02-17
feat: Forge support

hustlerone:
For now the first run **must** be connected to the internet.

Forge's Maven repo is cancer and literally withholds everything. Can I figure out the dependencies and vendor them? Yes. Is it cancer? Also yes.

Right now the current method is scraping. Surprise! They also don't give you hashes. Only sha1 or md5 which we don't give two fucks about.

As of writing the derivation is copypasted from Paper.


philocalyst:
Any update on what's needed from this? @Infinidoge ?

hustlerone:
Honestly this could be improved if we can find a database of hashes so we don't have to process all Forge jars.
Not that it takes a long time, but how things are right now it's kinda wasteful. I guess it's better than nothing, right?

hustlerone:
Also, as how things are right now, an internet connection is needed for the first launch. Technically speaking Forge supports creating offline installers but you still need the original .jar and i just wanted a minimally viable product

philocalyst:
Is that uncommon for a lot of these? Unfortunately I doubt there's a db like that around :(

▰▰▰▰▰
Miles Wirth 🙃

hustlerone:
> Is that uncommon for a lot of these? Unfortunately I doubt there's a db like that around :( ▰▰▰▰▰ Miles Wirth 🙃
> […](#)
> From: hustlerone ***@***.***> Sent: 27 April 2026 05:47 To: Infinidoge/nix-minecraft ***@***.***> Cc: Miles Wirht ***@***.***>, Comment ***@***.***> Subject: Re: [Infinidoge/nix-minecraft] feat: Forge support (PR #195) hustlerone left a comment ([Infinidoge/nix-minecraft#195](https://github.com/Infinidoge/nix-minecraft/pull/195)) Also, as how things are right now, an internet connection is needed for the first launch. -- Reply to this email directly or view it on GitHub: [#195 (comment)](https://github.com/Infinidoge/nix-minecraft/pull/195#issuecomment-4327074897) You are receiving this because you commented. Message ID: ***@***.***>

The other modloaders usually have an endpoint that contains the downloads, hashes and dependencies for their artefacts.

So far I've only been able to track down Forge's downloads.

philocalyst:
Still exciting!

Avelia-tr:
hey I tried to use this but it seems like pre 1.12.2 server just do not work ?
the config is just not applied on them and the datadir is basically emty but it does not show any error when building or anywhere that I could find ;s;

hustlerone:
> hey I tried to use this but it seems like pre 1.12.2 server just do not work ? the config is just not applied on them and the datadir is basically emty but it does not show any error when building or anywhere that I could find ;s;

Could you show logs and everything?

PR #200, opened 2026-03-16
feat: Add Folia support

viceebun:
Now that Folia has publicly downloadable builds, it seems easy enough to add Folia as a server option.
I've based it directly off ``paper-server``, and have added ``folia-server`` next to it.
The only tweak I've made was removing the JRE check, since Folia currently bases on 1.19 and above only.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -83,6 +83,7 @@
             quilt-server
             paper-server
             purpur-server
+            folia-server
             velocity-server
             minecraft-server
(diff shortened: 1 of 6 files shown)


jaredmontoya:
Note that you might have to migrate to https://fill.papermc.io/v3/projects/folia endpoint in the future like [paper has to](https://github.com/Infinidoge/nix-minecraft/pull/215)

PR #204, opened 2026-03-27
Feat/add folia support

tiaxter:


aaravrav:
why did you open this PR if one is already open?

tiaxter:
@aaravrav I didn't noticed there was another PR for the same reason

PR #230, opened 2026-07-13
fix(purpur-servers): use latest build

shunueda:
`listToAttrs` takes the first entry for duplicate keys. Since purpur's version only included the MC version (not the build number), all builds for the same MC version shared the same name, causing `listToAttrs` to always pick the oldest build. Fixed to match the behavior of paper server.


shunueda:
Hey @Infinidoge, could I give you a gentle bump on this?

